PR #54, opened 2019-04-24
S7/feature/custom user parameters

nicolocaruso85:


stu-greenshoots:
Can you add tests for this, please?

Test the public function updateUserFromUtterance passing in some different utterances and making sure the user is updated.

You will need to mock the getUser method to return a user without hitting DGraph. You can use something like this from Mockery: http://docs.mockery.io/en/latest/reference/creating_test_doubles.html#runtime-partial-test-doubles

UserServiceTest.php is added to the ignore list in phpunit.xml, so please add new tests in a separate test class so that they run on CI

PR #57, opened 2019-04-25
adds custom context support

stu-greenshoots:
OD-239

new test `CustomContextsTest` shows how it works. This is to support custom contexts that load their own attributes

### DIFF
--- a/src/ContextEngine/Contexts/Custom/AbstractCustomContext.php
+++ b/src/ContextEngine/Contexts/Custom/AbstractCustomContext.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace OpenDialogAi\ContextEngine\Contexts\Custom;
+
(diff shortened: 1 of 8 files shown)


src/ContextEngine/Contexts/Custom/AbstractCustomContext.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        parent::__construct(static::$name);
+    }
+
+    /**
+     * A function to load all custom attributes from any external sources into this custom context.
+     *
+     * All attributes should be added using @see AbstractContext::addAttribute()
+     */
+    abstract public function loadAttributes() :void;
```
stu-greenshoots: This might be easier as a function that returns an array of attributes that then get added as attributes by the context service...

PR #60, opened 2019-04-26
S7/feature/log messages

acrollet:
You'll need to add `<env name="LOCAL" value="1"/>` to your phpunit.xml and make sure dgraph is running to run the new tests...


src/ContextEngine/ContextParser.php
```diff
@@ -22,7 +22,7 @@ public static function determineContextAndAttributeId($attribute): array
         $contextId = null;
         $attributeId = null;
 
-        $matches = explode('.', $attribute);
+        $matches = explode('.', $attribute, 2);
```
acrollet: this is to deal with attribute names containing a '.' char

stu-greenshoots:
I think this is missing the step of creating the chatbot user... but that might be picked up but this ticket?
https://greenshootlabs.atlassian.net/browse/OD-178

src/ConversationLog/Service/ConversationLogService.php
```diff
@@ -0,0 +1,182 @@
...
+        return $intent->getLabel();
+    }
+
+    /**
+     * @param Intent $intent
+     * @return String
+     */
+    private function getSceneId(Intent $intent): string
+    {
+        return $this->userService->getSceneForIntent($intent->getUid());
```
stu-greenshoots: Is there a way to avoid going back to dgraph for this?
Could something be loaded into conversation contex? @istos

src/ContextEngine/ContextParser.php
stu-greenshoots: If we allow attribute names with a . then we won't be able to tell if it's an attribute name, or a context.attribute name

PR #61, opened 2019-04-26
Add request/response logging

acrollet:


stu-greenshoots:
I don't think we need the util space to be a separate package, we can just consider it part of core. The migrations can live in the top level migrations folder

routes/web.php
```diff
@@ -1,13 +1,20 @@
...
     Route::get('/config', function () {
         return config('opendialog.core');
-    });
+    })->middleware(RequestLoggerMiddleware::class);
 
-    Route::post('/incoming/webchat', 'OpenDialogAi\SensorEngine\Http\Controllers\WebchatIncomingController@receive');
+    Route::post(
+        '/incoming/webchat',
+        'OpenDialogAi\SensorEngine\Http\Controllers\WebchatIncomingController@receive'
+    )->middleware(RequestLoggerMiddleware::class);
```
stu-greenshoots: Is there a benefit to having this registered as a global middleware? 
https://laravel.com/docs/5.8/middleware#global-middleware

src/Util/Http/Middleware/RequestLoggerMiddleware.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        $requestLength = microtime(true) - LARAVEL_START;
+        $memoryUsage = memory_get_usage();
+
+        RequestLog::create([
+            'url' => $request->url(),
+            'query_params' => serialize($request->query()),
+            'method' => $request->method(),
+            'source_ip' => $request->ip(),
+            // FIXME get this.
```
stu-greenshoots: let's get this in.

I suggest:
- look at the logger being used in odai-webchat and move that into core
- make a change so that the uid for a user is generated/got from the service later
- The value of it can be stored using the core service provider (perhaps binding a primitive like described here https://laravel.com/docs/5.8/container#binding)
- The end goal is that anything that wants to grab the request id will get the same value

stu-greenshoots:
Might just be my set up, but I can't seem to run the migrations. I'm sure moving this logic into core rather than a utils package will solve that though

stu-greenshoots:
Updating this to:
- split the response headers and content into separate fields
- json encode rather than serialise (we want to be able to read these raw, json is a better format)

stu-greenshoots:
And added a switch in the env file to stop logging (on by default)

PR #68, opened 2019-05-03
OD-210: Remove conversation from DGraph on “unpublish” action

nicolocaruso85:


istos:
@nicolocaruso85 going to suggest a different way of doing this that puts in some of the groundwork to handle versioning of conversations. 

1. In the MySQL Conversation model of conversation builder add a new field called graph_uid

2. When a conversation is published to DGraph update the Laravel model to hold the graph_uid

3. When a conversation is requested to be unpublished from dgraph:
a. Do a Dgraph query to find the conversation by name - for this write a new query in ConversationQueryFactory called getConversationUid. 
b. Confirm that the uid returned matches the one we have on file - 
if not then raise an alert but don't delete the conversation. 
if yes delete the conversation

PR #86, opened 2019-05-14
S8/feature/webchat formatter update

acrollet:


stu-greenshoots:
I pulled the main parse message function apart into smaller ones to make it clearer to read and pulled out all literals into constants

PR #87, opened 2019-05-14
Trim message elements when formatting

acrollet:
OD-270

### DIFF
--- a/src/ResponseEngine/Message/WebchatMessageFormatter.php
+++ b/src/ResponseEngine/Message/WebchatMessageFormatter.php
@@ -58,8 +58,8 @@ class WebChatMessageFormatter implements MessageFormatterInterface
                 $messages[] = $this->parseMessage($item);
             }
 
-            if (isset($message['disable_text'])) {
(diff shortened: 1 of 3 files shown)


src/ResponseEngine/Message/WebchatMessageFormatter.php
```diff
@@ -241,7 +241,7 @@ protected function getMessageText(SimpleXMLElement $element): string
                         }
 
                         if ($link['url']) {
-                            $text .= $this->generateLink($link['url'], $link['text'], $link['open-new-tab']);
+                            $text .= ' ' . $this->generateLink($link['url'], $link['text'], $link['open-new-tab']);
```
acrollet: @stu-greenshoots just alerting you to this - not sure if we'd rather do this or special-case the trimming to allow trailing white space in message text when there's a link... c.f. testWebChatTextMessageWithLink in ResponseEngineTest.php

stu-greenshoots: I'm fine with this for now.

PR #90, opened 2019-05-15
S8/bug/delete existing message templates

acrollet:
I did some general cleanup as well.

### DIFF
--- a/src/Console/Commands/ExportConversation.php
+++ b/src/Console/Commands/ExportConversation.php
@@ -40,7 +40,7 @@ class ExportConversation extends Command
             exit;
         }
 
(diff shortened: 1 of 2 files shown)


src/Console/Commands/ExportConversation.php
```diff
@@ -14,7 +14,7 @@ class ExportConversation extends Command
      *
      * @var string
      */
-    protected $signature = 'export:conversation {conversation name} {--f|filename=}';
+    protected $signature = 'conversation:export {conversation name} {--f|filename=}';
```
stu-greenshoots: yeah, that's nicer :)

PR #104, opened 2019-06-05
OD-310: Add support for messages of type “RichMessage”

nicolocaruso85:


stu-greenshoots:
Can you:
- add validation to MessageXML
- update validation in MessageXML to have a default case that returns false

PR #105, opened 2019-06-05
OD-309: Add support for messages of type “ListMessage”

nicolocaruso85:


stu-greenshoots:
Can you update message.xml rule to validate this message type

PR #108, opened 2019-06-05
S8/feature/run allthetests

stu-greenshoots:
*You will need to override the DGraph URL locally to be able to run tests*
Create a .env file with DGRAPH_URL=http://10.0.2.2 in it, otherwise it will use the default of localhost

### DIFF
--- a/src/ConversationBuilder/Conversation.php
+++ b/src/ConversationBuilder/Conversation.php
@@ -171,10 +171,11 @@ class Conversation extends Model
      *
      * @param ConversationNode $conversation
      * @return bool
+     * @throws \Illuminate\Contracts\Container\BindingResolutionException
      */
     public function publishConversation(ConversationNode $conversation)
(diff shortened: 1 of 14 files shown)


src/ConversationBuilder/Conversation.php
```diff
@@ -171,10 +171,11 @@ public function buildConversation()
      *
      * @param ConversationNode $conversation
      * @return bool
+     * @throws \Illuminate\Contracts\Container\BindingResolutionException
      */
     public function publishConversation(ConversationNode $conversation)
     {
-        $dGraph = new DGraphClient(env('DGRAPH_URL'), env('DGRAPH_PORT'));
+        $dGraph = app()->make(DGraphClient::class);
```
stu-greenshoots: I was going round in circles for a while trying to figure out why tests were failing locally... this line is why!

PR #147, opened 2019-06-26
Bug/message conditions

stu-greenshoots:


src/ResponseEngine/Service/ResponseEngineService.php
```diff
@@ -119,26 +119,21 @@ protected function messageMeetsConditions(MessageTemplate $messageTemplate): boo
...
 
                 $attribute = $this->getAttributeForCondition($attributeName, $contextId);
-                $conditionsPass = $condition->compareAgainst($attribute);
-            }
-        }
 
-        if ($conditionsPass) {
-            return true;
+                if (!$condition->compareAgainst($attribute)) {
+                    $conditionsPass = false;
```
stu-greenshoots: Maybe just return false here so we don' have to continue

src/ResponseEngine/Service/ResponseEngineService.php
```diff
@@ -119,26 +119,21 @@ protected function messageMeetsConditions(MessageTemplate $messageTemplate): boo
...
+                }
+            }
         }
 
-        return false;
+        return $conditionsPass;
```
stu-greenshoots: and just return true here?

PR #149, opened 2019-06-28
Feature/performance improvements

stu-greenshoots:


stu-greenshoots:
I need to fix the broken tests in this PR that are still trying to use the conversation/chatbot user together

Issue #169, opened 2019-07-22
Message XML with ampersand (&)


stu-greenshoots:
If a message template XML contains an ampersand symbol, it always fails validation unless it is escaped (converted to &amp;)

stu-greenshoots:
PR: https://github.com/opendialogai/core/pull/183

PR #183, opened 2019-08-07
Escape ampersands in Message XML

nicolocaruso85:
fixes https://github.com/opendialogai/core/issues/169


stu-greenshoots:
I think this issue is more far-reaching than just in the validator - I will review where else we need to be conscious of this and update this PR

stu-greenshoots:
@istos - I think this one needs some deeper thought to get to the crux of the issue

stu-greenshoots:
Happy to merge now to prevent a stale PR and will review issue #169 to make sure it has been addressed

Issue #125, opened 2019-06-17
Should mapping a callback ID to an intent be required?


stu-greenshoots:
The callback interpreter is used as the default interpreter and will map a callback id from a button/trigger etc message to an intent id.

If the callback ID is not mapped to an intent, the interpreter will not match an intent.

When designing a conversation, it's not always clear what callback id has been mapped to an intent, and since conversation design uses intents, but buttons in messages should use callback it's easy to lose track and not see why conversations are not matching

stu-greenshoots:
Update the callback interpreter in the following way:
Check if the callback Id is in the supported callbacks. If it is, return the corresponding intent value. 

If not, return an intent with the value of the callback

stu-greenshoots:
https://github.com/opendialogai/core/pull/185

PR #185, opened 2019-08-07
Update the callback interpreter to return an intent with the value of the callback if the callback Id is not in the supported callbacks

nicolocaruso85:


nicolocaruso85:
fixes https://github.com/opendialogai/core/issues/125

stu-greenshoots:
This is using `$utterance->getValue()` it should be using `$utterance->getCallbackId()`

I will update

stu-greenshoots:
@nicolocaruso85 I have updated this to use the callback_id instead of the value as it was.

Can you please add some tests to the App/Interpreter module to test the callback interpreter on it's own.

Issue #114, opened 2019-06-12
Nothing is added to the logs when publishing/unpublishing a conversation 


stu-greenshoots:
Maybe not in all cases, but when I try to unpublish a conversation, nothing is added to the laravel log

stu-greenshoots:
https://github.com/opendialogai/core/pull/187

stu-greenshoots:
I will close this issue for now as it will be handled as part of https://github.com/opendialogai/core/issues/194

PR #187, opened 2019-08-07
Unpublish a conversation when the name is changed

nicolocaruso85:
Related to #194


src/ConversationBuilder/Observers/ConversationObserver.php
```diff
@@ -20,14 +22,29 @@ public function created(Conversation $conversation)
...
-    public function updated(Conversation $conversation)
+    public function updating(Conversation $conversation)
     {
-        //
+        // Unpublish the conversation from DGraph.
+        if ($conversation->isDirty('name')) {
+            if ($conversation->getOriginal('status') == 'published') {
+                $originalName = $conversation->getOriginal('name');
+
+                $dGraph = new DGraphClient(env('DGRAPH_URL'), env('DGRAPH_PORT'));
```
stu-greenshoots: Don't instantiate DGraphClient like this - it should be pulled from the service layer

stu-greenshoots:
Would be even better for the DGraph dependency to be injected into the Conversation builder observer if possible, or if not, loaded globally in a constructor mether

nicolocaruso85:
fixes https://github.com/opendialogai/core/issues/114

stu-greenshoots:
Closing this as it should be picked up as part of https://github.com/opendialogai/core/issues/194

PR #189, opened 2019-08-13
Updates getNextPossibleBotIntents to handle more complex conversations

### LINKED ISSUE #188: Unexpected filtering behaviour when determining the next possible outgoing intents
When determining the next intent, the `ConversationEngine.getNextIntent` method calls `getNextPossibleBotIntents` on the current scene object. In this method, all outgoing intents are filtered. As seen below, intents are retained if they are said across scenes or if they are in uninterrupted ascending order.

https://github.com/opendialogai/core/blob/fb979f0608393544b128768d35e6d23d30dff857/src/Conversation/Scene.php#L145-L157

This requirement of the order being uninterrupted causes a constraint on how conversation markup is written. In the two screenshots below, the difference in the highlighted areas determines whether the next intent will be correctly filtered even though the markup appears to be equivalent.

The conversation markup when not working as expected

Using the above markup, with the incoming intent as `intent.shoebot.sign_up_for_updates`, the `intent.shoebot.ask_name` intent is filtered out when you would expect it to be retained. This is because its order is two greater than `intent.shoebot.signup_for_updates` rather than one (which is required on line 152 above). 

Currently to achieve the expected behaviour, the following markup is required to ensure the filter functions in the expected way.

The conversation markup required currently to fix the issue

elliotmassen:
This PR fixes #188 by updating `Scene.getNextPossibleBotIntents`. Methods have also been added to the `Scene` and `Participant` classes to get intents sorted by their orders. `getNextPossibleBotIntents` has been updated to use these new methods, filter through the ordered bot intents and remove those that are lower than the current intent's order and which are not within a sequential block after the current intent.

The decision of which intent to choose is still left with the ConversationEngine. `getNextPossibleBotIntents` will return an ordered map containing either the sole next bot intent or a block of sequential next bot intents. Tests have also been added to check and demonstrate this functionality.

### DIFF
--- a/src/Conversation/Participant.php
+++ b/src/Conversation/Participant.php
@@ -69,6 +69,16 @@ class Participant extends Node
     /**
      * @return \Ds\Map
      */
+    public function getAllIntentsSaidInOrder()
+    {
+        return $this->getAllIntentsSaid()->sorted(function (Node $a, Node $b) {
+            return $a->getAttributeValue(Model::ORDER) > $b->getAttributeValue(Model::ORDER);
+        });
+    }
+
+    /**
+     * @return \Ds\Map
+     */
     public function getAllIntentsListenedFor()
     {
         $allIntentsSaid = new Map();
--- a/src/Conversation/Scene.php
+++ b/src/Conversation/Scene.php
@@ -152,26 +152,20 @@ class Scene extends NodeWithConditions
      */
     public function getNextPossibleBotIntents(Intent $currentIntent): Map
     {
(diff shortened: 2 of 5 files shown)


stu-greenshoots:
My only comment would be to reduce the need for lengthy comments in your own code. It is nice and clean and readable as it is, so the comments can actually sometimes prove distracting to the flow.

I think you are right to put in a similar method for users as for bots even though this PR isn't using it, it is a useful future addition.

PR #190, opened 2019-08-14
Adds support for simple and most prebuilt LUIS entity types

elliotmassen:
This PR furthers LUIS entity type support from just list entities to include [simple entities](https://docs.microsoft.com/en-us/azure/cognitive-services/luis/reference-entity-simple) and most of the [prebuilt entity types](https://docs.microsoft.com/en-us/azure/cognitive-services/luis/luis-concept-entity-types#prebuilt-entity). The fallback for these few unsupported prebuilt types is to resolve the entity using the entity string (which is also the case for simple entity types).

The updates anticipate three potential structures of an entity object within a LUIS response:

1. When no `resolution` property exists, get the entity value from the `entity` property.
2. When a `resolution` property exists and it contains a `values` property, get the entity value from the first item in the list.
3. When a `resolution` property exists and it contains a `value` property, get the entity value from there.

### DIFF
--- a/src/InterpreterEngine/Luis/LuisEntity.php
+++ b/src/InterpreterEngine/Luis/LuisEntity.php
@@ -4,6 +4,10 @@ namespace OpenDialogAi\InterpreterEngine\Luis;
 
 class LuisEntity
 {
+    private const NO_RESOLUTION_VALUE = 0;
+    private const ONE_RESOLUTION_VALUE = 1;
+    private const MANY_RESOLUTION_VALUES = 2;
+
     /* @var string */
     private $type;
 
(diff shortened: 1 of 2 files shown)


src/InterpreterEngine/Luis/LuisEntity.php
```diff
@@ -110,8 +112,41 @@ public function getScore()
...
+        foreach ($values as $value) {
             $this->resolutionValues[] = $value;
         }
     }
+
+    /**
+     * @param $entity
+     * @return int
+     */
+    private function detectResolutionStructure($entity): int
```
elliotmassen: @stu-greenshoots Do you think that using constants is correct here?

stu-greenshoots: Yes, I like the use of constants here. 
I would just add a small description to the phpdoc of this method

PR #191, opened 2019-08-15
Extract entity value from the query string for simple entities

elliotmassen:
This PR amends the LuisEntity class to extract the entity value from the query string for simple (or unknown) LUIS entities. This allows simple LUIS entities to retain any user-inputted cases, as LUIS converts all input to lowercase.

### DIFF
--- a/src/InterpreterEngine/Luis/LuisEntity.php
+++ b/src/InterpreterEngine/Luis/LuisEntity.php
@@ -8,6 +8,12 @@ class LuisEntity
     private const ONE_RESOLUTION_VALUE = 1;
     private const MANY_RESOLUTION_VALUES = 2;
 
+    /**
+     * The user's input
+     * @var string
+     */
+    private $query;
+
     /* @var string */
     private $type;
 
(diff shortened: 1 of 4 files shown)


src/InterpreterEngine/Luis/LuisEntity.php
```diff
@@ -150,4 +164,20 @@ private function detectResolutionStructure($entity): int
 
         return self::NO_RESOLUTION_VALUE;
     }
+
+    /**
+     * Extracts a value when the entity has no resolution value. The value is extracted from the the query if possible
+     * and from the entity string otherwise.
+     * @return string
+     */
+    private function extractValueWhenNoResolution()
```
stu-greenshoots: Just for consistency, can you add the return type here and on any methods that don't define one (ie `extractValueWhenNoResolution(): string`)

stu-greenshoots:
Changes look good to me and are covered by unit tests 👍 

1 code style comment so we can start building some consistency

PR #192, opened 2019-08-19
Only return opening intents when searching for a new conversation

elliotmassen:
This PR updates the `AllOpeningIntents.getIntents` method to return only the first user intents from opening scenes. Prior to this PR, the method returns user intents from throughout the opening scene, which is not the desired behaviour.

Returning these additional intents creates the possibility of a user dropping into the middle of a different conversation template instead of being switched to the No Match conversation. An instance when this could occur is if the user enters text and one of the incorrectly returned intents has a Luis Interpreter - the user's text may well match this intent and set it to be the next intent, therefore dropping the user into a new conversation late in its opening scene.

This PR also fixes SceneTest which incorrectly relied on this previous behaviour.

### DIFF
--- a/src/ConversationEngine/ConversationStore/DGraphQueries/AllOpeningIntents.php
+++ b/src/ConversationEngine/ConversationStore/DGraphQueries/AllOpeningIntents.php
@@ -5,6 +5,7 @@ namespace OpenDialogAi\ConversationEngine\ConversationStore\DGraphQueries;
 
 
 use Ds\Map;
+use Ds\Set;
 use OpenDialogAi\Core\Conversation\Model;
 use OpenDialogAi\Core\Graph\DGraph\DGraphClient;
 use OpenDialogAi\Core\Graph\DGraph\DGraphQuery;
@@ -95,7 +96,7 @@ class AllOpeningIntents extends DGraphQuery
 
             if (isset($conversation[Model::HAS_OPENING_SCENE])) {
                 if (isset($conversation[Model::HAS_OPENING_SCENE][0][Model::HAS_USER_PARTICIPANT])) {
-                    $matchedIntents = $this->extractIntentsFromParticipant(
+                    $matchedIntents = $this->extractOpeningIntentsFromParticipant(
                         $conversation[MODEL::HAS_OPENING_SCENE][0][Model::HAS_USER_PARTICIPANT][0]
                     );
                     foreach ($matchedIntents as $intent) {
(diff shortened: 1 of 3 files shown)


src/ConversationEngine/ConversationStore/DGraphQueries/AllOpeningIntents.php
```diff
@@ -153,18 +131,49 @@ public function getIntents()
...
+            $intentsArray = array_merge($intentsArray, $participant[Model::SAYS]);
         }
 
         if (isset($participant[Model::SAYS_ACROSS_SCENES])) {
-            return $participant[Model::SAYS_ACROSS_SCENES];
+            $intentsArray = array_merge($intentsArray, $participant[Model::SAYS_ACROSS_SCENES]);
+        }
+
+        // We need to use the array before the Map to ensure that there are no issues with duplicate keys
+        for ($i = 0; $i < count($intentsArray); $i++) {
```
stu-greenshoots: Let's try using DS\Set or DS\Stack to avoid the need for an array and a Map.

src/ConversationEngine/ConversationStore/DGraphQueries/AllOpeningIntents.php
```diff
@@ -153,18 +131,49 @@ public function getIntents()
      * Extract intents from a participant in an opening scene. Looks for says or says_across_scenes relationships
      *
      * @param $participant
-     * @return array
+     * @return Map
      */
-    private function extractIntentsFromParticipant($participant)
+    private function extractIntentsFromParticipant($participant): Map
     {
+        $intents = new Map();
```
istos: You could probably just do a foreach on Model::SAYS and Model::SAYS_ACROSS_SCENES and place the entries in a Map with an index that you manage yourself (same way you do in the for loop). That would save you from having to create arrays. Probably a bit less work computationally although overall not a huge difference I would suspect.

src/ConversationEngine/ConversationStore/DGraphQueries/AllOpeningIntents.php
```diff
@@ -153,18 +131,49 @@ public function getIntents()
...
+        // We need to use the array before the Map to ensure that there are no issues with duplicate keys
+        for ($i = 0; $i < count($intentsArray); $i++) {
+            $intents->put($i, $intentsArray[$i]);
         }
 
-        return [];
+        $previousKeptIntent = null;
+
+        // Sort the intents by order and then filter them so that we get just the first user intent(s)
+        $intents = $intents->sorted(
```
istos: Good stuff!

src/ConversationEngine/ConversationStore/DGraphQueries/AllOpeningIntents.php
stu-greenshoots: This has been updated to use a Set - no index required

src/ConversationEngine/ConversationStore/DGraphQueries/AllOpeningIntents.php
```diff
@@ -65,12 +65,12 @@ public function __construct(DGraphClient $client)
             ]);
 
         $response = $client->query($this);
-        $this->data = $response->getData();
+        $this->conversations = $response->getData();
```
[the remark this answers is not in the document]
elliotmassen: It seems like we should always get back a set of conversation templates or an empty array (if there are none). Considering how proximate the query definition is, I'm not clear on how much validation would be necessary. Would you be able to elaborate?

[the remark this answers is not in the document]
istos: We might also get an exception from the client - https://github.com/opendialogai/core/blob/07079afaf526956e9a15fd9be0a7fcec2bcac21f/src/Graph/DGraph/DGraphClient.php#L130 - which would lead to a break since we are not handling it here. However, this is a wider issue since we are not handling exceptions in most places where we do queries like this! As such let us park that issue for this particular PR - fine for me to merge.

Issue #172, opened 2019-07-23
Invalid characters in dgraph queries


stu-greenshoots:
When trying to save a conversation that contains a star (*) the triple mutation fails with the following error:

    0 => array:2 [▼
      "code" => "ErrorInvalidRequest"
      "message" => "while lexing _:a_level_confirm <has_expected_attribute> _:user.a* .: Invalid character '*' found for itemType: 8"
    ]

OpenDialogAi\Core\Graph\DGraph\DGraphMutationResponse __construct
…/vendor/opendialogai/core/src/Graph/DGraph/DGraphClient.php81

nicolocaruso85:
https://github.com/opendialogai/core/pull/195

PR #195, opened 2019-08-23
Escapes non-valid characters from the dgraph mutation

nicolocaruso85:


nicolocaruso85:
fixes https://github.com/opendialogai/core/issues/172

Issue #196, opened 2019-08-27
Update InterpreterService to cache interpreter results for the same Utterance


stu-greenshoots:
The same utterance can be run through the same interpreter multiple times if the interpreter is used in many opening intents.

We should cache the results for the same utterance and interpreter in the InterpreterService when it first runs.

Then if the interpreter is called again with exactly the same utterance for the same interpreter, the cached result for the intent array should be returned.

Update the interpreter() method in the InterpreterService (and interface) to first check if there is a cached value, and if not use the interpreter to get the intent array, and cache that value. 

+ Put info level log messages when retrieving a value from the cache to show that it is a cached value.
+ Make the cache time configurable per interpreter
  + Update the opendialog-interpreterengine-custom file and add a new config array named `interpreter_cache_times`. This should be a key-value pair array of 'interpreter_name' => cache time in seconds
 + Add a 'default_cache_time' value to the config. The interpreter service should use this for any interpreter that does not have a defined cache time
+ Use separate methods to get the cache time for the interpreter, cache values and retrieve cached values
+ Use Laravel default caching mechanisms - we can select the appropriate driver later

PR #197, opened 2019-08-27
Update InterpreterService to cache interpreter results for the same Utterance

nicolocaruso85:


nicolocaruso85:
fixes https://github.com/opendialogai/core/issues/196

Issue #155, opened 2019-07-10
If an action is not bound, the conversation fails


stu-greenshoots:
We should probably:
a) inform the conversation designer if they are using any actions that have not been bound
b) fail in a more graceful manner

stu-greenshoots:
@elliotmassen can you pick this up, please?

For now, let's make sure we catch and log any exception from trying to get a nonexistent action, but make sure that the conversation continues

PR #198, opened 2019-08-27
Catches exception that was thrown when an non-binded action was calle…

elliotmassen:
This PR is related to #155. This update catches the expectation that is thrown when a non-binded action is called. Rather than breaking the conversation, this is now logged as a warning.


stu-greenshoots:
fixes https://github.com/opendialogai/core/issues/155

Issue #140, opened 2019-06-17
Message Templates should favour those with passing conditions


stu-greenshoots:
When finding the right message template for an outgoing intent, we should favour those that have conditions that have been met over those that have no conditions.

EG if there is a message that has the condition user.name is set and a message with not conditions, if the user has a name, we should pick the message with the user.name is set condition.

1. Select all messages with passing conditions
2. Order messages based on how many conditions they have that passed (more conditions push them to the top)
3. Pick the message with the most conditions matched.

PR #204, opened 2019-09-04
When there are multiple message templates pick the message with the most conditions matched

nicolocaruso85:


nicolocaruso85:
fixes https://github.com/opendialogai/core/issues/140

Issue #150, opened 2019-07-02
Exporting conversations keeps the conversation id


stu-greenshoots:
Exporting conversations is exporting the conversation id along with the conversation. This won't make sense when the conversation is re-imported

stu-greenshoots:
This refers to the `conversation:export` command

PR #205, opened 2019-09-04
Improve conversations Import and Export

nicolocaruso85:


nicolocaruso85:
fixes https://github.com/opendialogai/core/issues/150

Issue #199, opened 2019-09-02
Use DGraph docker compose from OD github repo


stu-greenshoots:
The dgraph docker-compose file included with OD should not be stored in the repo, it should use https://github.com/opendialogai/dgraph-docker so that updates can be handled centrally for any place we use DGraph.

We should update `CoreServiceProvider` to move dgraph docker-compose from the vendor folder instead and delete the local docker-compose

PR #207, opened 2019-09-04
Use DGraph docker compose from OD github repo

nicolocaruso85:


nicolocaruso85:
fixes https://github.com/opendialogai/core/issues/199

PR #209, opened 2019-09-06
Separation of concerns in ConversationStore

### LINKED ISSUE #201: Separation of concerns in ConversationStore
## General

When discussing the changes suggested in #194 with @stu-greenshoots & @istos, issues with the current implementation of the Conversation Engine's `DGraphConversationStore` were highlighted. The summary of these issues is that there does not currently exist a separation of concerns between `DGraphQueryFactory`, `DGraphClient` and the returned data. This means that additions and updates to graph queries (as will be required by #194) are not necessarily applied in similar ways.

This issue proposes to address this by clearly separating the three components. While `DGraphQueryFactory` and `DGraphClient` already exist, there is not a uniform way to return queried data. "EI Models" will be added to begin introducing some consistency in returned data. These models will allow us to define expected data structures eg. `EIModelOpeningIntents`. Alongside this addition, `DGraphConversationStore` and `DGraphQueryFactory` will need to be restructured to accommodate the new separation of concerns.

## Example snippet
The following is an example method of `DGraphConversationStore` (provided by @istos).

```php
function getAllOpeningIntents(): EIModelAllOpeningIntents
{
	/*@var DGraphQuery $query */
	$query = DGraphQueryFactory::getAllOpeningIntents();
	
	$response = $this->client->dGraphClient->query($query);
​
	/*@var EIModelIntents $eiOpeningIntents */
	$eiOpeningIntents = $this->eiModelCreator->createEIModel(eiModelCreator.OpeningIntents, $response);
​
	return $eiOpeningIntents;
}
```

### ConversationStore/DGraphConversationStore

*   The methods of `DGraphConversationStore` will need to be updated to match the above example snippet:
    *   Rather than passing the `DGraphClient` instance to each `DGraphQueryFactory` method, the Conversation Store will now retrieve the query from the factory and pass it to the client itself.
    *   The response will then be passed to `EIModelCreator` to retrieve an EI Model.
    *   Each method will return an EI Model.

### ConversationStore/DGraphQueries/ConversationQueryFactory

*   Rename “ConversationQueryFactory” to “DGraphConversationQueryFactory”, create a “ConversationQueryFactory” interface that it implements and bind the concrete to the interface in the ConversationEngine service provider.
*   Calling the `DGraphQuery.query` method will be removed from the factory. Instead it will now return the object so that the ConversationStore can run it.
*   Remove `buildConversationFromDGraphData` (and related methods) as these will be superseded by EI Model’s. 
    *   It is worth noting that `UserSerice.getCurrentConversation` & `ConversationStore.getCurrentConversation` relies on this, so we will need to ensure that the newly returned EI Model is converted into a Conversation/Conversation object. This can be easily achieved by moving the `buildConversationFromDGraph` method elsewhere and replacing array indexing with accessor class to the EI Model.

### Graph/DGraph/DGraphClient

*   `DGraphClient` seems to meet the requirements already, no changes should be required to this component.

### ConversationStore/EIModelCreator

*   The `EIModelCreator.createEIModel` method of the creator will take a creator const (to signify the type of model to be created) and a `DGraphQueryResponse` object and will return an EI Model object.
*   Depending on the given const, a handler method will be selected to return the desired model.
*   Each handler method will parse the response object and marshal it into the desired model.
*   For now, each model will follow the naming convention of the class name beginning with “EIModel”. In the future there may be reason to create an `EIModel` base class, but for now this does not seem necessary.

elliotmassen:
This PR resolves #201 by refactoring the ConversationStore's methods to clearly separate query management, the `DGraphClient` and representing the data. Each method in the query factory now returns a `DgraphQuery` object, rather than handle any querying or conversion. These objects are then passed to the `DGraphClient`, which remains unchanged by this PR. Responses from the client (`DGraphQueryResponse`'s) are now turned into various `EIModel`'s, using the new `EIModelCreator`.

Each `EIModel` specifies static `validate` and `handle` methods, which respectively allow us to confirm that the response contains the necessary data and to create an instance of the model. A variety of models have been specified in this PR, with the possibility of more being added later if required.

This change has also necessitated creating an `EIModelConversationConverter` class to allow for conversion from `EIModelConversation` objects to `Conversation` objects. At a later date we may wish to make this converter redundant by reducing the need for an in-memory representation of the graph, but this is outside the scope of #201. The methods within the converter are a port of the (now deleted) ConversationStore methods, such as `buildConversationFromDGraphData`.


src/ContextEngine/Contexts/User/UserService.php
```diff
@@ -299,7 +306,7 @@ public function userIsHavingConversation($userId): bool
     /**
      * @param $userId
      * @return mixed
```
istos: Not strictly about this PR, but does this really return`mixed` or would it be a Conversation.

src/ConversationEngine/ConversationEngine.php
```diff
@@ -86,9 +86,10 @@ public function setActionEngine(ActionEngineInterface $actionEngine): void
      * @param UserContext $userContext
      * @param UtteranceInterface $utterance
      * @return Intent
+     * @throws FieldNotSupported
      * @throws GuzzleException
      * @throws NodeDoesNotExistException
-     * @throws FieldNotSupported
+     * @throws ConversationStore\EIModelCreatorException
```
istos: Minor point but we should either have full path or no path in these definitions, in general I would opt for no path as with the other exceptions.

src/ContextEngine/Contexts/User/UserService.php
elliotmassen: Ah yes, this should be a `Conversation` object being returned. This is now updated.

src/ConversationEngine/ConversationEngine.php
elliotmassen: This is now updated.

Issue #202, opened 2019-09-02
Only the user context is persisted after adding intent attributes


stu-greenshoots:
See the `TODO` here: https://github.com/opendialogai/core/blob/develop/src/ConversationEngine/ConversationEngine.php#L431

Rather than just tracking whether the user context has been updated, the `storeIntentAttributes` method should keep track of all contexts being updated.

The `ContextInterface` should be updated to include a persist method that each context needs to implement (in the case where a context does not need to be persisted, it can do nothing, or just create a log message)

Update the user context to use this persist method.

PR #212, opened 2019-09-09
Add “persist” method to all contexts and update “storeIntentAttributes” to call persist on all contexts

nicolocaruso85:
fixes https://github.com/opendialogai/core/issues/202


src/ConversationEngine/ConversationEngine.php
```diff
@@ -422,11 +419,13 @@ private function storeIntentAttributes(Intent $intent, Map $expectedAttributes =
...
             Log::debug(sprintf('Storing attribute %s in %s context', $attribute->getId(), $context->getId()));
             $context->addAttribute($attribute);
+
+            $contextsUpdated[$context->getId()] = $context->getId();
         }
 
-        // TODO - is there a better way of doing this? Each context could have it's own tear down method to deal with persisting
-        if ($userContextUpdated) {
-            ContextService::getUserContext()->updateUser();
+        foreach ($contextsUpdated as $contextId) {
```
stu-greenshoots: Can you pull this into a separate method 'persistContexts(array $contexts)

src/ContextEngine/ContextManager/ContextInterface.php
```diff
@@ -52,4 +52,10 @@ public function addAttribute(AttributeInterface $attribute);
      * @return bool true if removed, false if not
      */
     public function removeAttribute(string $attributeName): bool;
+
+    /**
+     * Persist context value.
+     * In the case where a context does not need to be persisted, it can do nothing, or just create a log message.
+     */
+    public function persist();
```
stu-greenshoots: Can you make this return a bool - true if the context was persisted successfully

PR #214, opened 2019-09-10
Feature/sensor interface

JustSteveKing:
Extended core to follow: #PP-64 on Jira.

Webchat Interfaces have now been created and implementations now extend, service provider bindings are in place. Providing better abstraction on how OpenDialog handles incoming messages, allowing future extension into more than just HTTP.

Inclusion of grumPHP to monitor and manage standards, with 2 default tasks:
- PHPCS (PSR12)
- PHPCS Fixer 

Addition on `composer.json` file to add scripts -> test for quick access to testing tools for all users.


src/SensorEngine/SensorEngineServiceProvider.php
```diff
@@ -27,5 +30,7 @@ public function register()
             $mysqlLinkClick = new MySqlLinkClick();
             return $mysqlLinkClick;
         });
+
+        $this->app->bind(IncomingMessageInterface::class, IncomingWebchatMessage::class);
```
stu-greenshoots: This binding will make less sense when we
a) move webchat items out to a separate package
b) introduce the XMPP package

Happy for it to live here until those changes start coming in though

stu-greenshoots:
Cool, happy with the work in here. Would have been easier to check if the refactoring and grumPHP were in a separate PR :)

Issue #222, opened 2019-09-16
Logging Updates


stu-greenshoots:
1. Move the logging (that logs user id and request id) into the core package. Make any fixes / updates required.

2. Install the monolog introspection processor (https://github.com/Seldaek/monolog/blob/master/src/Monolog/Processor/IntrospectionProcessor.php) to log more info with each log line.
(see here for more info on logging the right file)

PR #225, opened 2019-09-16
Add IntrospectionProcessor

nicolocaruso85:
fixes https://github.com/opendialogai/core/issues/222

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -71,6 +71,11 @@ All queries are logged at info level
 By default, all incoming and outgoing API calls will be logged to the request and response mysql tables.
 To prevent this happening, set the `LOG_API_REQUESTS` env variable to `false`
 
+## Introspection logging
+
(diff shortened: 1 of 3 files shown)


stu-greenshoots:
This now produces log lines like this:
```[2019-09-19 21:37:17] testing.DEBUG: Got a matching conversation for user 2574fea0-dc69-33ea-bf85-f5754a4245b3 with id no_match_conversation  {"request_id":"od-5d83f50c3e2ea7.48667566","file":"/home/vagrant/code/OpenDialogAi-Core/src/ConversationEngine/ConversationEngine.php","line":154,"class":"OpenDialogAi\\ConversationEngine\\ConversationEngine","function":"determineCurrentConversation"}```

I think it might be a bit to much to log all the time.

I think the Introspection logger should only be added if the log level is DEBUG (not only added to DEBUG level messages, only add the helper to the logging stack if the env for LOG_LEVEL = DEBUG)

PR #229, opened 2019-10-16
Better Conversation Handling between App and Dgraph backend

### LINKED ISSUE #193: When setting the current conversation, outdated versions of conversation templates are retrieved
When a conversation template is updated and re-published, the previous version of the template is left unaltered in DGraph (related to #157). This can cause unnecessary iterations in `ConversationEngine.setCurrentConversation` when it looks for opening intents, as the query used still retrieves these outdated versions.

Storing these previous versions seems to serve a purpose - @stu-greenshoots suggested they are useful if a user is having a conversation while it is updated - however there doesn't exist any form of garbage collection or any way in DGraph to differentiate between current and previous versions of conversation templates. If we wanted to differentiate currently, we need to pull the `graph_uid`'s from the MySQL `conversations` table and use them in a GraphQL+- filter.

At the moment it seems like repercussions from this issue are avoided by exporting and re-importing conversations to clear out DGraph, but a more permanent solution will help avoid any potential issues when numerous changes are made in production.

### LINKED ISSUE #194: Better Conversation Handling between App and Dgraph backend
The integration between the admin of conversations and Dgraph is very rudimentary and can easily lead to inconsistencies between the state of conversations in the app admin and Dgraph. 

Related Issues:
#193 
#157 
#114 

We need to define the desired behaviour and work towards that.

elliotmassen:
This PR resolves #194 and is dependent on the changes from #209.

#### Definitions
A key change is to conversation statuses. Previously statuses were `imported`, `invalid`, `validated` & `published`, these have now been changed to `saved`, `activatable`, `activated`, `deactivated` and `archived`.

The `saved` status is a combination of the previous `imported` & `invalid` statuses, denoting a conversation that has not yet been validated. After successful validation a conversation becomes `activatable`, which succeeds the previous `validated` status.

The `activated` & `deactivated` statuses ensure that we can specifically query activated conversations when looking looking for opening intents, which fixes #193 and ensures that we are only ever offering the intended conversations to a user. When a new template version is activated, the previous version will be automatically deactivated which ensures that there will only ever exist one activated version of a template.

The `archived` status gives a way to hide unwanted, deactivated conversations. Like all other non-activated statuses, archived conversations will not be considered when querying for opening intents. Archived conversations can either be completely deleted from both MySQL and DGraph databases or be restored as a deactivated conversation.

#### Transitions
As laid out in @istos' [diagram](https://github.com/opendialogai/core/issues/194#issuecomment-524314232), there are defined transitions between statuses. These transitions are as follows: 

| Current status | Potential following statuses |
| --- | --- |
| `saved` | `activatable` |
| `activatable` | `saved`, `activated` |
| `activated` |  `saved`, `deactivated` |
| `deactivated` | `activated`, `archived` |
| `archived` | `deactivated`, **permanently deleted** |

These transitions are enforced by the new methods in `ConversationManager`, each throwing a `InvalidConversationStatusTransitionException` if the conversation is not in a currently in correct status to move to the desired status.

https://github.com/opendialogai/core/blob/b511966d43f71447786a7de926c5250d461b2562/src/Conversation/ConversationManager.php#L275-L296

#### Relationships
Another addition in this PR are the new graph relationships, `update_of` & `instance_of`. The `update_of` relationship connects two conversation templates and denotes that the 'from' node is an updated version of the 'to' node. The `instance_of` relationship connects a conversation template and a user's conversation and denotes that the user's conversation is a instantiation of the template.

#### Attributes
The `update_of` relationship provides the basis for versioning of conversation templates and to support this a `conversation_version` attribute has been added to conversation templates in the graph. Starting at zero, the version number increments with each successive version of the template.

A new `conversation_status` attribute has also been added to track the new status updates mentioned above.

### Version history
To track and display the version history of each conversation, Spatie's [ActivityLog](https://docs.spatie.be/laravel-activitylog/v3/introduction/) has been used. Changes to a conversation's model, notes, status, version number or graph UID cause a snapshot of the conversation to be logged. It's worth noting that we only increment the version number when a conversation has been activated (i.e has been re-persisted to the graph). The OpenDialog application can then query the ActivityLog and filter on changes that incremented the version number to get a full version history of a conversation template (see more in opendialogai/opendialog#72).

#### Multiple filters
Prior to this PR, our DGraph support only allowed for a single query filter using the `eq` function. This PR expands the support to include multiple filters (connected by `and` or `or` connectives). Each query filter allows use of all currently supported functions (`eq`, `uid`, `has` & `allofterms`) as well as some negations (`not eq` & `not has`).

#### Recurse
To allow deletion of a conversation template and its history, support for recurse queries has been added. Using a recurse query on the new `update_of` relationship, we can discover all previous template versions. Recurse queries can be created as follows:

```php
$query = (new DGraphQuery())->uid("0xTEST")->recurse()->setQueryGraph(
    Model::UID,
    Model::UPDATE_OF
);
```

This produces a query such as the following:

```qraphql
query {
    node(func: uid(0xTEST)) @recurse(loop:false) {
        uid
        update_of
    }
}
```

#### Expanded use of Schema types
Following the introduction of a `User` type in #206, the use of types has now been expanded to include `Conversation`, `Scene`, `Participant` & `Intent`. These are defined in the graph schema specified in `DGraphClient`. This ensures that we are able to use DGraph's `expand(_all_)` on nodes of these types. It also provides the option to phase out use of the `ei_type` attribute in favour of using the built-in types, which may be of interest considering the introduction of a [`type` query function](https://docs.dgraph.io/query-language/#using-types-during-queries) in DGraph. 

#### Final considerations
A future consideration for support in our DGraph integration would be sorting and basic pagination. When building a conversation for persisting, this PR adds code that fetches the latest version of the template for the graph so that the `update_of` relation can be properly added. However without support for sorting or pagination, all versions of the template are queried and the latest version is found in PHP as seen in the snippet below.

It would be more efficient (and remove the possibility of a large version history slowing down OpenDialog) if our DGraph integration support sorting and at least use of DGraph's `first` syntax, so that this logic could all be wrapped up in the query.

https://github.com/opendialogai/core/blob/b511966d43f71447786a7de926c5250d461b2562/src/ConversationEngine/ConversationStore/DGraphConversationStore.php#L137-L154


src/ConversationEngine/ConversationEngine.php
```diff
@@ -269,13 +269,17 @@ private function setCurrentConversation(UserContext $userContext, UtteranceInter
...
+        $conversationForCloning = $this->conversationStore->getConversation($intent->getConversationUid(), true);
+
+        /** @var Conversation $conversationForConnecting */
+        $conversationForConnecting = $this->conversationStore->getConversation($intent->getConversationUid(), false);
 
         // TODO can we avoid building, cloning and re-persisting the conversation here. EG clone directly in DGRAPH
         // TODO and store the resulting ID against the user
 
-        $userContext->setCurrentConversation($conversation);
+        $userContext->setCurrentConversation($conversationForCloning, $conversationForConnecting);
```
istos: We could do the relationship the other way round so that we don't need a conversationForConnecting, the conversationForCloning object can carry the "is_instance_of" relationship with the UID to the right conversation template.

src/ConversationEngine/ConversationStore/DGraphConversationStore.php
```diff
@@ -89,6 +91,68 @@ public function getEIModelConversationTemplate($conversationTemplateName): EIMod
...
+    }
+
+    /**
+     * @param $templateName
+     * @return EIModelConversation
+     * @throws EIModelCreatorException
+     */
+    public function getLatestEIModelTemplateVersionByName($templateName): EIModelConversation
+    {
+        $query = DGraphConversationQueryFactory::getConversationFromDGraphWithTemplateName($templateName);
```
istos: How much extra work would it be to supporting sorting in Dgraph itself.

src/Graph/DGraph/DGraphClient.php
```diff
@@ -121,15 +130,15 @@ public function alter(string $alter)
 
         try {
             return $this->getData($response);
-        } catch (\Exception $e) {
+        } catch (Exception $e) {
```
istos: Can we have a dedicated exception for this.

src/ConversationEngine/ConversationStore/DGraphConversationStore.php
elliotmassen: I've just taken another look at the syntax for sorting and getting the first X results in DGraph, and I'm sure it won't be any longer than a couple of hours to add support for them. Let me know if you'd like support added as part of the feedback.

src/Conversation/ConversationManager.php
```diff
@@ -205,4 +226,110 @@ public function botSaysToUserAcrossScenes($startingSceneId, $endingSceneId, Inte
 
         return $this;
     }
+
+    /**
+     * Sets the conversation status to 'activatable' if it is currently 'saved'
+     * @throws InvalidConversationStatusTransitionException
+     */
+    public function setValidated(): void
```
stu-greenshoots: Can the logic be consolidated here?
Set up a static array of valid states in the form:
    static $states = [
        Conversation::DEACTIVATED => [{VALID_STATE_1},{VALID_STATE_2}],
        Conversation::ACTIVATED => [{VALID_STATE_1},{VALID_STATE_2}],
    ];

Then have a `private function setState($fromState, $toState)` that wraps up checking if the to state is valid from the from state, setting the status and formatting the error log message if not

src/ConversationBuilder/Conversation.php
```diff
@@ -59,19 +58,19 @@ class Conversation extends Model
         'notes',
         'created_at',
         'updated_at',
+        'version_number'
```
stu-greenshoots: Can you add this as an annotated property of the class (nice for type hinting and auto-completion)

src/ConversationBuilder/Conversation.php
```diff
@@ -218,43 +245,173 @@ public function publishConversation(ConversationNode $conversation)
...
+        return true;
+    }
+
+    /**
+     * Deactivate the conversation in DGraph.
      * @return bool
+     * @throws \Illuminate\Contracts\Container\BindingResolutionException
      */
-    public function unPublishConversation($reValidate = true)
+    public function deactivateConversation(): bool
```
stu-greenshoots: Again, can we consolidate this logic please

src/ConversationEngine/ConversationEngine.php
stu-greenshoots: My question here is can we avoid making 2 calls out to DGraph to get a conversation? Can we get it once and strip the UIDS after the fact?

src/ConversationEngine/ConversationStore/DGraphConversationQueryFactory.php
```diff
@@ -153,11 +166,26 @@ public static function getIntentByUid(string $intentUid): DGraphQuery
      * @return array
      */
     public static function getConversationQueryGraph(): array
+    {
+        return array_merge(
+            self::getConversationTemplateQueryGraph(),
+            [
+                Model::INSTANCE_OF => self::getConversationTemplateQueryGraph()
```
stu-greenshoots: Is this `INSTANCE_OF`, or `UPDATE_OF`?

stu-greenshoots:
Code coverage has gone down (very very slightly) - but I think that is mainly due to the new commands. I don't think we need tests for the commands though 👍

src/ConversationEngine/ConversationStore/DGraphConversationQueryFactory.php
elliotmassen: I believe this should be `INSTANCE_OF` because when dealing with a user conversation (which the above is a query graph for) we don't have an `UPDATE_OF` relationship.

src/ConversationEngine/ConversationStore/ConversationStoreInterface.php
```diff
@@ -37,6 +37,31 @@ public function getConversation($conversationId, bool $clone = false): Conversat
...
+     * @return EIModelConversation
+     * @throws EIModelCreatorException
+     */
+    public function getEIModelConversationTemplateByUid($uid): EIModelConversation;
+
+    /**
+     * @param $templateName
+     * @return Conversation
+     */
+    public function getLatestTemplateVersionByName($templateName): Conversation;
```
istos: No need to change this now @elliotmassen , but just as a note for future work I would avoid functions that are actually parametrised by they function name :-) 

`get<Order>TemplateVersionByName($templateName)` should probably be `getTemplateVersionByName($templateName, $versionType`

src/ConversationEngine/ConversationStore/DGraphConversationStore.php
istos: Yes, I think let's go ahead and support that then.

src/Graph/DGraph/DGraphClient.php
```diff
@@ -23,6 +24,13 @@ class DGraphClient
     const DELETE = 'delete';
     const MUTATE_COMMIT_NOW = 'mutate?commitNow=true';
 
+    const INTENT = 'Intent';
+    const PARTICIPANT = 'Participant';
+    const SCENE = 'Scene';
+    const CONVERSATION = 'Conversation';
+    const USER = 'User';
```
istos: Since these are OD specific terms rather than Dgraph terms they should really be here. Worth a discussion to understand where they are being used.

src/ConversationEngine/ConversationStore/DGraphConversationQueryFactory.php
stu-greenshoots: ok, I'm a bit confused then. What makes this query graph a user conversation one?

stu-greenshoots: We agreed to update the function name

stu-greenshoots:
I can't find the right bit of code to comment on now, but we discussed moving the logic of the `delete()` function in ConversationBuilder/Conversation into the ConverstionObserver @elliotmassen

src/ConversationBuilder/Conversation.php
```diff
@@ -218,43 +245,173 @@ public function publishConversation(ConversationNode $conversation)
...
+    }
+
+    /**
+     * @return bool
+     * @throws \Illuminate\Contracts\Container\BindingResolutionException
+     * @throws \OpenDialogAi\ConversationEngine\ConversationStore\EIModelCreatorException
+     * @throws Exception
+     * @throws \GuzzleHttp\Exception\GuzzleException
+     */
+    public function delete(): bool
```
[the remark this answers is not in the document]
elliotmassen: This is now updated as per the second paragraph of your comment.

src/ConversationEngine/ConversationEngine.php
elliotmassen: Just to update on this, I had a chat with @istos and we agreed to leave this as is for the moment. For future reference, this solution means we only have to do one DGraph mutation (as when persisting the ChatbotUser to the graph, the user conversation is picked up via the `having_conversation` relationship and then the template is picked up via the `instance_of` relationship). However the downside is that we are carrying around another large in-memory conversation object to achieve this.

src/Graph/DGraph/DGraphClient.php
elliotmassen: These constants are only used in the DGraph schema and in DGraph mutation preparation, where they are used to map `ei_type`'s to DGraph types. Perhaps they could be moved into a SchemaService when #234 is resolved?

src/ConversationEngine/ConversationStore/DGraphConversationStore.php
elliotmassen: This has now been added.

PR #232, opened 2019-10-21
Attribute handling

### LINKED ISSUE #220: DGraph User type does not include predicates for custom attributes
In #206 DGraph support was upgraded to support v1.1.0. A key change is the introduction of types and OpenDialog now defines a `User` type. Types affect the behaviour of `expand(_all_)` as per the [DGraph documentation](https://docs.dgraph.io/query-language/#expand-predicates):

> When expand(\_all_) is called on this node, Dgraph will first check which types the node has. Then it will query the definitions of the types and build a list.

This causes an issue when `ConversationEngine.storeIntentAttributes` is run as this calls the `persist` method of `UserContext` which in turn calls `UserService.updateUser`. This method persists user attributes to the graph, then re-queries the graph and returns the data. The returned data is used to update the `UserContext`'s `user` property. This re-query of the graph uses `expand(_all_)`, which means that we would need to have all our custom attributes defined in the schema for the `User` type, which we do not.

This means that the custom attributes are persisted in the graph, but because the query does not return them, the application thinks they do not exist.

As seen in the snippet below, currently we only define type predicates for supported attributes, **not custom attributes**.

A solution to this would be to use the `AttributeResolver::getSupportedAttributes` method, however this does not escape the fact that custom attributes must be defined in the graph schema now if we wish to use `expand(_all_)` to query user predicates. Bearing in mind that the schema is only initialised when `artisan conversations:setup` is run, this limits the ease of adding new custom attributes (as you need to re-initialise the graph schema after any additions).

https://github.com/opendialogai/core/blob/1065e421502d8cdfc075a7b5916d98689c3c47a3/src/Graph/DGraph/DGraphClient.php#L269-L305

elliotmassen:
This PR fixes #220 by changing the way we store custom attributes in the graph. Previously custom attributes were stored in the graph as values, each defined by an edge named after the attribute ID. Now the `User` graph type has a `has_attribute` predicate from which multiple `UserAttribute` nodes can be connected - each node contains values for the attribute ID, type & value.

### New class: UserAttribute
This PR adds a new `GraphItem` class called `UserAttribute`, which represents a custom attribute within the User context. Each `UserAttribute` node object is constructed by passing the `AttributeInterface` object from OpenDialog which is stored internally so that it can be accessed later.

### UserContext
The various methods implemented by `HasAttributesTrait` on `AbstractContext` are still overridden in this PR. However, while previously they accessed the `ChatbotUser` object's attributes directly, they now go through the `UserService` to access the new `UserAttributes` on the `ChatbotUser`.

### ChatbotUser
`ChatbotUser` has a set of new methods for manipulating `UserAttributes` that are connected to it, such as `addUserAttribute`, `getUserAttribute` & `getAllUserAttributes`. Behind the scenes these manipulate relationships between the `ChatbotUser` node and various `UserAttribute` nodes. It's worth noting that `ChatbotUser` still makes use of the `HasAttributeTrait` as it still has some standard attributes that should be stored as values (such as id & ei_type).


src/ContextEngine/Contexts/User/UserContext.php
```diff
@@ -82,7 +95,7 @@ public function setAttribute(string $attributeName, $value, $type = null): Attri
      */
     public function removeAttribute(string $attributeName): bool
     {
-        return $this->getUser()->removeAttribute($attributeName);
+        return false;
```
stu-greenshoots: Should this do something? How do we remove an attribute from a user?

src/ContextEngine/Contexts/User/UserContext.php
```diff
@@ -33,45 +36,55 @@ public function __construct(
...
 
     /**
      * @inheritDoc
      */
-    public function addAttribute(AttributeInterface $attribute)
+    public function getAttribute(string $attributeName): AttributeInterface
     {
-        $this->getUser()->addAttribute($attribute);
+        /** @var UserAttribute $userAttribute */
+        $userAttribute = $this->getAttributes()->get($attributeName);
```
stu-greenshoots: What happens here if the attribute does not exist?

stu-greenshoots:
Discussed 2 things with Elliot:
- The naming of UserAttribute. As attributes might start existing against things other than users, the name is too specific. I think just using `Attribute` will be OK in this instance

- The fact that the user context works a little differently to the other contexts now and doesn't extend AbstractContext. We discussed converting the attributes into the more generic OD attribute usage (ie having a map of attributes that can be manipulated with the HasAttributesTrait) when the context is loaded (and in the reverse when the context is persisted)

stu-greenshoots:
Now I am looking again, the User context already overrode lots of the HasAttributesTrait - lets not worry about this until we have a better grasp on other contexts

src/ContextEngine/Contexts/User/UserContext.php
elliotmassen: This has now been updated to mirror functionality from `HasAttributeTrait`.

src/ContextEngine/Contexts/User/UserContext.php
elliotmassen: This has now been updated to mirror functionality from `HasAttributeTrait`.

elliotmassen:
It's worth noting that this PR allows us to increase test efficiency by removing the newly introduced initialisation of DGraph for every test case.

src/ContextEngine/Contexts/User/UserContext.php
elliotmassen: @istos Ah yes I see what you mean. The has & remove methods make use of the `getAttributes` method (directly & indirectly respectively) which does call the `UserService`, so on a functionality level there is consistency, but from a readability level it initially appears not to be the case.I could have them all call the `UserService` directly, but then we wouldn't have any need for

src/ContextEngine/Contexts/User/UserService.php
```diff
@@ -365,15 +340,13 @@ public function unsetCurrentConversation(ChatbotUser $user): void
     }
 
     /**
-     * @param $userId
      * @return DGraphQuery
      */
-    private function getUserQuery($userId): DGraphQuery
+    public function getUserQuery(): DGraphQuery
```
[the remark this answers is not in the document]
elliotmassen: Ah yes that's a much better method name, thanks!

src/ContextEngine/Contexts/User/UserService.php
```diff
@@ -393,6 +366,15 @@ private function getUserQuery($userId): DGraphQuery
         return $dGraphQuery;
     }
 
+    /**
+     * @param $userId
+     * @return DGraphQuery
+     */
+    public function getSingleUserQuery($userId)
```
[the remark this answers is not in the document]
elliotmassen: Indeed 👍

PR #235, opened 2019-10-25
Intent conditions

elliotmassen:
This PR adds conditions to outgoing intents and is depends on #229. It is related to opendialogai/opendialog#79.

### Outgoing intent conditions
Conversation now support conditions on outgoing intents. The markup is exactly the same as conversation conditions and are handled by the `OperationService` in exactly the same way. An example of conditions on outgoing intents is as follows.

```yaml
conversation:
  id: with_outgoing_intent_conditions
  scenes:
    opening_scene:
      intents:
        - u: 
            i: hello_bot
        - b: 
            i: response_with_name
            conditions:
                - condition:
                    operation: is_set
                    attributes:
                        attribute1: user.user_name
            completes: true
        - b:
            i: response_without_name
            conditions:
                - condition:
                    operation: is_not_set
                    attributes:
                        attribute1: user.user_name
            completes: true
```

### Multiple opening intents
Additionally, this PR also adds support for multiple opening incoming intents in an opening scene as the scope of this PR initially included conditions on incoming intents as well. The scope was reduced however changes to the conversation model have been retained as they may be useful in the future (and don't seem to cause any harm). The `opening_intent` attribute has now been changed to `opening_intents` and returns an array of strings rather than a string. This change is also accounted for in the OpenDialog application as seen in opendialogai/opendialog#79.


src/Controllers/OpenDialogController.php
```diff
@@ -56,6 +60,10 @@ public function setResponseEngine(ResponseEngineServiceInterface $responseEngine
      * @param UtteranceInterface $utterance
      * @return OpenDialogMessages
      * @throws FieldNotSupported
+     * @throws GuzzleException
```
stu-greenshoots: This guzzle exception is coming all the way up from DGraph client. I have created an issue in core to look into DGraph error handling - we should probably deal with the exception at a much lower level

src/ContextEngine/Contexts/User/UserContext.php
```diff
@@ -203,27 +203,21 @@ public function hasCurrentIntent(): bool
...
-     * @throws \GuzzleHttp\Exception\GuzzleException
-     * @throws \OpenDialogAi\Core\Graph\Node\NodeDoesNotExistException
+     * @throws CurrentIntentNotSetException
      */
     public function getCurrentScene(): Scene
     {
-        if ($this->user->hasCurrentIntent()) {
-            $currentIntent = $this->conversationStore->getEIModelIntentByUid($this->user->getCurrentIntentUid());
+        if (!$this->user->hasCurrentIntent()) {
+            throw new CurrentIntentNotSetException("Attempted to get the current scene without having set a current intent.");
```
stu-greenshoots: I have created a separate issue for discussion around Exceptions not being caught in the flow. Something to ponder on. https://github.com/opendialogai/core/issues/258

src/ContextEngine/Contexts/User/UserContext.php
```diff
@@ -203,27 +203,21 @@ public function hasCurrentIntent(): bool
...
+        $currentIntent = $this->conversationStore->getEIModelIntentByUid($this->user->getCurrentIntentUid());
 
-            // use the conversation that is against the user
-            $currentScene = $this->userService->getCurrentConversation($this->user->getId())->getScene($sceneId);
-        } else {
-            // Set the current intent as the first intent of the opening scene
-            /* @var Scene $currentScene */
-            $currentScene = $this->user->getCurrentConversation()->getOpeningScenes()->first()->value;
+        // Get the scene for the current intent
+        $sceneId = $this->userService->getSceneForIntent($currentIntent->getIntentUid());
```
stu-greenshoots: I have created an issue for the separation of concerns between user queries and general conversation queries:
https://github.com/opendialogai/core/issues/259

PR #239, opened 2019-10-29
Scene conditions

elliotmassen:
This PR adds the ability to specify scene conditions. It is dependent on #235.

### Scene conditions
Conditions can be added to scenes in the following way:

```yaml
conversation:
  id: with_scene_conditions
  scenes:
    opening_scene:
      intents:
        - u:
            i: opening_user_s1
            scene: scene1
        - u:
            i: opening_user_none
        - b: 
            i: opening_bot_response
            completes: true
    scene1:
      conditions:
        - condition:
            operation: is_not_set
            attributes:
              attribute1: user.user_email
      intents:
        - b: 
            i: scene1_bot_response
            completes: true
```

#### Functionality
In terms of functionality, the scene conditions will be pulled from scenes onto the respective cross-scene intents ("said across scenes"). This happens in two places, for opening intents in `EIModelOpeningIntents.collectSceneIntents` and for regular intents in `Intent.getConditionsFromDestinationScene`. In both cases the graph is followed from the cross-scene intent to the participant that listens for it and then up to the scene that the participant is within. From here the conditions are retrieved and added alongside the intent (& potential conversation conditions) for checking.

Previously this PR attempted to use the `ConversationManager` to achieve this. The scene conditions were retrieved and added when calling `userSaysToBotAcrossScenes` or `botSaysToUserAcrossScenes`. While this worked, it meant that the scene conditions were persisted against the intents in the graph, and would be re-added each time, causing consistency issues (as well as memory issues).

#### Behaviour
In the example conversation given above, if the bot receives a 'opening_user_s1' intent, it will have access to the scene1's conditions and these will be checked. If the conditions pass the conversation will progress to scene1, if they fails it will result in a 'NoMatchResponse'. In the future, if we were to allow multiple occurrences of the same intent and we were to have a sequential block of the same intent pointing to various scenes, each destination scenes' conditions would be checked and the conversation would progress to the first scene with passing conditions.

### DIFF
--- a/src/Conversation/Intent.php
+++ b/src/Conversation/Intent.php
@@ -309,15 +309,15 @@ class Intent extends NodeWithConditions
      */
     public function getAllConditions(): Map
     {
-        $intents = new Map();
+        $conditions = new Map();
 
         if ($this->hasConditions()) {
-            $intents = $intents->merge($this->getConditions());
+            $conditions = $conditions->merge($this->getConditions());
         }
 
-        $intents = $intents->merge($this->getConditionsFromDestinationScene());
+        $conditions = $conditions->merge($this->getConditionsFromDestinationScene());
 
-        return $intents;
+        return $conditions;
     }
 
     /**
--- a/src/Conversation/Intent.php
+++ b/src/Conversation/Intent.php
@@ -303,4 +303,48 @@ class Intent extends NodeWithConditions
             $this->addAttribute($attribute);
         }
     }
+
+    /**
+     * @return Map
+     */
+    public function getAllConditions(): Map
+    {
+        $intents = new Map();
+
+        if ($this->hasConditions()) {
+            $intents = $intents->merge($this->getConditions());
+        }
+
+        $intents = $intents->merge($this->getConditionsFromDestinationScene());
+
+        return $intents;
+    }
+
+    /**
+     * @return Map
+     */
+    private function getConditionsFromDestinationScene(): Map
+    {
+        $conditions = new Map();
+
+        if ($this->hasIncomingEdgeWithRelationship(Model::LISTENS_FOR_ACROSS_SCENES)) {
+            /** @var Participant $participant */
+            $participant = $this->getNodesConnectedByIncomingRelationship(Model::LISTENS_FOR_ACROSS_SCENES)->first()->value;
+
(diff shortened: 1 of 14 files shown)


src/ConversationEngine/ConversationStore/EIModels/EIModelOpeningIntents.php
```diff
@@ -38,27 +41,16 @@ public static function validate(array $response, $additionalParameter = null): b
      * @param array|array $response
      * @param null $additionalParameter
      * @return EIModel
-     * @throws \Illuminate\Contracts\Container\BindingResolutionException
-     * @throws \Exception
+     * @throws BindingResolutionException
+     * @throws Exception
```
istos: Is this still required?

src/ConversationEngine/ConversationStore/EIModels/EIModelOpeningIntents.php
```diff
@@ -7,7 +7,10 @@
 use Countable;
 use Ds\Map;
 use Ds\Set;
+use Exception;
```
istos: Are we actually using Exceptions?

src/Conversation/Intent.php
```diff
@@ -303,4 +303,48 @@ public function copyNonCoreAttributes(Intent $intent): void
             $this->addAttribute($attribute);
         }
     }
+
+    /**
+     * @return Map
+     */
+    public function getAllConditions(): Map
+    {
+        $intents = new Map();
```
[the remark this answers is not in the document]
elliotmassen: Ah yes, this should've been `$intents` - this is now updated.

src/ConversationEngine/ConversationStore/EIModels/EIModelOpeningIntents.php
elliotmassen: These were not required, I've now updated the method to only throw a `EIModelCreatorException`.

src/ConversationEngine/ConversationStore/EIModels/EIModelOpeningIntents.php
elliotmassen: After the recent update we are not. This has now been removed.

PR #243, opened 2019-10-31
Feature/odc x  local test env

madan95:


.lando.yml
```diff
@@ -0,0 +1,59 @@
+name: opendialog-core
+recipe: laravel
```
stu-greenshoots: Instead of the Laravel recipe (which includes nginx, node and mysql), can you see if we can just use the php service directly:

https://docs.lando.dev/config/php.html#patch-versions

config/php.ini
```diff
@@ -0,0 +1,6 @@
+xdebug.remote_enable=1
```
stu-greenshoots: I _think_ we can get away without this as long as phpstorm is set up in the correct way

stu-greenshoots:
When I run this locally, it looks like it creates a php:7.3-apache-2 container:
image

was expecting php:7.3-fpm

.lando.yml
```diff
@@ -0,0 +1,57 @@
...
+  dgraph-alpha:
+  - dgraph-alpha.lndo.site:8000
+services:
+  php-service:
+    type: php:7.3
+    via: cli
+    webroot: .
+    xdebug: true
+    config:
+      php: config/php.ini
```
stu-greenshoots: Do we need this?

stu-greenshoots:
This still brings up an apache flavoured php, but happy to stick with that for the moment

Issue #238, opened 2019-10-29
Integrate with Rasa NLU


istos:
We would like to be able to use the RASA NLU service for intent resolution. 

There is a repository setup with a simple to use Rasa container - https://github.com/opendialogai/rasa-nlu-development - and some basic instructions on how to train a RASA model. 

We want an interpreter in the style of the LUIS Interpreter - https://github.com/opendialogai/core/tree/develop/src/InterpreterEngine/Luis

---
Using Rasa through Lando

Place the following:
```
## intent:initiate_search
- search for information on London
- what do you have on lettuce
- give me information on Obanma
- I am looking for a guide to Paris

## intent:ask_for_help
- how does this work
- help
- provide assistance
- help me
```
In a file called nlu.md in the directory data and then run `lando train`

you should get a model in the models directory which you can then start with 
`lando serve-model "models/name"`

Hitting localhost on port 5005 with
```
{
	"text":"give me info on London"
}
```

should produce

```
{
    "intent": {
        "name": "initiate_search",
        "confidence": 0.8180559683
    },
    "entities": [
        {
            "entity": "GPE",
            "value": "London",
            "start": 16,
            "confidence": null,
            "end": 22,
            "extractor": "SpacyEntityExtractor"
        }
    ],
    "intent_ranking": [
        {
            "name": "initiate_search",
            "confidence": 0.8180559683
        },
        {
            "name": "ask_for_help",
            "confidence": 0.1819440317
        }
    ],
    "text": "give me info on London"
}
```
This is the response that our OD interpreter can pickup and deal with.

PR #244, opened 2019-10-31
Rasa Interpreter

nicolocaruso85:


nicolocaruso85:
fixes https://github.com/opendialogai/core/issues/238

src/InterpreterEngine/config/opendialog-interpreterengine.php
```diff
@@ -19,8 +19,13 @@
         'endpoint_key' => env('QNA_ENDPOINT_KEY'),
     ],
 
+    'rasa_config' => [
+        'app_url'      => env('RASA_APP_URL'),
```
stu-greenshoots: *  Not directly related to this PR *
Does the RASA API offer any kind of security? If not, we should make sure the port is not open to the wider world when we deploy

stu-greenshoots:
Good work - reviewed the review and the updates from @elliotmassen - happy with all of it.
I pushed in some very minor tweaks (mainly removing blank lines) and added a comment around RASA security.

Happy for you to merge this @elliotmassen

PR #253, opened 2019-11-08
Support cross-scene intents with destination as opening scene

nicolocaruso85:


elliotmassen:
Thanks for this Nicolo! While drafting feedback for this PR, I realised the problem was caused by a previously existing issue that I have since resolved in another open PR (#248). Therefore this ticket will now be resolved as part of that. It turns out the main aspect of this bug was relating to `Scene.getNextPossibleUserIntents`! No further action needed from you on this ticket. I'll now close this PR.

Issue #250, opened 2019-11-07
Conversation->activateConversation() should not need the conversation passed in


stu-greenshoots:
In ConversationBuild/Conversation, the activate method takes a conversation node to activate. In all cases I have seen, this ends up with the rather circular code of:

```
$conversation->activateConversation($conversation->buildConversation());
```

Let's just make `activateConversation()` call `buildConversation()`

PR #254, opened 2019-11-08
Call function “buildConversation” inside “activateConversation”

nicolocaruso85:
fixes https://github.com/opendialogai/core/issues/250

### DIFF
--- a/src/ConversationBuilder/Conversation.php
+++ b/src/ConversationBuilder/Conversation.php
@@ -209,13 +209,15 @@ class Conversation extends Model
     /**
      * Activate the conversation in DGraph.
      *
-     * @param ConversationNode $conversation
+     * @param Conversation $conversation
      * @return bool
      * @throws \Illuminate\Contracts\Container\BindingResolutionException
      */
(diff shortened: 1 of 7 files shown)


src/ConversationBuilder/Conversation.php
```diff
@@ -209,13 +209,15 @@ public function buildConversation()
...
-     * @param ConversationNode $conversation
+     * @param Conversation $conversation
      * @return bool
      * @throws \Illuminate\Contracts\Container\BindingResolutionException
      */
-    public function activateConversation(ConversationNode $conversation): bool
+    public function activateConversation(Conversation $conversation): bool
     {
-        $cm = ConversationManager::createManagerForExistingConversation($conversation);
+        $conversationNode = $conversation->buildConversation();
```
elliotmassen: This method shouldn't take any parameters and instead  `$this->buildConversation()` should be called to get the conversation node.

elliotmassen:
Thanks Nicolo, this looks good to me - this just requires the the feedback in opendialogai/opendialog#87 to be resolved now.

PR #260, opened 2019-11-12
PP-197 intial implementation of composite attribute and collection

madan95:


tests/Feature/CompositeAttributeTest.php
```diff
@@ -0,0 +1,47 @@
...
+class CompositeAttributeTest extends TestCase
+{
+    public function testCompositeAttribute()
+    {
+        // createFromInput
+        $attributeCollection = new ExampleAbstractAttributeCollection(
+            array(['id' => 'one', 'value' => 'go']),
+            'array'
+        );
+        $compositeAttribute = new ExampleAbstractCompositeAttribute(
```
stu-greenshoots: Can you add 'c' to the registered attributes and call AttributeResolver::getAttributeFor('c', $attributeCollection);

stu-greenshoots:
I'm happy with this approach. I think it will develop as we work further on the CompositeAttribute type, but this fits the bill for now 👍

PR #265, opened 2019-11-13
Removed depricated context praser helper method & php cs fixes

madan95:


src/ContextEngine/ContextParser.php
```diff
@@ -21,7 +21,7 @@ abstract class ContextParser
     public static function determineContextAndAttributeId($attribute): array
```
stu-greenshoots: Let's delete this method as it shouldn't be used anymore

src/ContextEngine/ContextParser.php
```diff
@@ -21,7 +21,7 @@ abstract class ContextParser
...
     {
         $parsedAttribute = self::parseAttributeName($attribute);
-        return [$parsedAttribute->context, $parsedAttribute->attributeId];
+        return [$parsedAttribute->contextId, $parsedAttribute->attributeId];
```
stu-greenshoots: I think we now have `$context` and `$contextId` in parsed attribute - let's just stick with `$contextId`

src/ContextEngine/ParsedAttributeName.php
```diff
@@ -13,6 +13,10 @@ class ParsedAttributeName
 
     public $attributeId = AbstractAttribute::INVALID_ATTRIBUTE_NAME;
 
+    public $contextId = AbstractAttribute::UNDEFINED_CONTEXT;
+
+    public $accessor = [];
+
     public $itemNumber;
```
stu-greenshoots: $itemNumber and $itemName should be removed (along with any getters or setters)

src/ContextEngine/ParsedAttributeName.php
```diff
@@ -13,6 +13,10 @@ class ParsedAttributeName
 
     public $attributeId = AbstractAttribute::INVALID_ATTRIBUTE_NAME;
 
+    public $contextId = AbstractAttribute::UNDEFINED_CONTEXT;
```
stu-greenshoots: Why do we have contextId and context? They are the same thing here

src/OperationEngine/Service/OperationService.php
```diff
@@ -84,7 +84,9 @@ public function checkCondition(Condition $condition)
         $attributes = [];
 
         foreach ($condition->getOperationAttributes() as $name => $attribute) {
-            [$contextId, $attributeName] = ContextParser::determineContextAndAttributeId($attribute);
+            $parsedAttributeName = ContextParser::parseAttributeName($attribute);
+            $contextId = $parsedAttributeName->contextId;
```
stu-greenshoots: Just use these directly in the code below - ie don't create a new variable

src/ResponseEngine/Formatters/Webchat/WebchatMessageFormatter.php
```diff
@@ -305,7 +305,10 @@ public function generateTextMessage(array $template): OpenDialogMessage
      */
     protected function getAttributeMessageText($attributeName): string
     {
-        [$contextId, $attributeId] = ContextParser::determineContextAndAttributeId($attributeName);
+        $parsedAttributeName = ContextParser::parseAttributeName($attributeName);
+        $contextId = $parsedAttributeName->contextId;
```
stu-greenshoots: Same here, just use $parsedAttributeName->contextId directly and don't create the $contextId variable

src/ContextEngine/tests/ContextParserTest.php
```diff
@@ -53,22 +53,22 @@ public function testArrayNotationAttribute()
...
         $attribute = "user.test[1]";
 
-        $this->assertEquals('user', ContextParser::parseAttributeName($attribute)->context);
+        $this->assertEquals('user', ContextParser::parseAttributeName($attribute)->contextId);
         $this->assertEquals('test', ContextParser::parseAttributeName($attribute)->attributeId);
-        $this->assertEquals(1, ContextParser::parseAttributeName($attribute)->itemNumber);
+        $this->assertEquals(1, ContextParser::parseAttributeName($attribute)->getAccessor()[0]);
         $this->assertNull(ContextParser::parseAttributeName($attribute)->itemName);
 
         // We do not support this depth
```
stu-greenshoots: We do support this depth now :)

So let's get some tests in. 

We need to remove any checks for itemNumber or itemName (as they will be removed from the ParsedAttributeName)

In this test, user.test[1][test][test] should create a parsed attribute name with:

$contextId = user
$attributeId = test
$accessor = [1, test, test]

PR #268, opened 2019-11-19
Supporting multiple outgoing intent

elliotmassen:
This PR adds support for multiple outgoing intents to the OpenDialogController. The purpose of adding this support is to eventually enable virtual incoming intents, which will cause multiple outgoing intents to be returned for a single request.

### ConversationEngine 
`ConversationEngine.getNextIntent` has become `ConversationEngine.getNextIntents` and now returns an array of intents. For the moment it only returns a single intent in this array but subsequent PR's (as mentioned above) will add the possibility of two or more intents being returned.

### `next_intent` attribute
The `next_intent` attribute is now an `ArrayAttribute`. Various test methods have been updated to reflect this.

### Message model
The `Message` model (used for tracking incoming & outgoing messages) has had its `intent` attribute renamed to `intents`. This now stores an array of intent `id`'s, rather than a single intent `id`. A migration has also been created to update the database table.


istos:
This all looks good - as long as we resolve the conflict happy for this to be merged.

src/Controllers/OpenDialogController.php
```diff
@@ -109,4 +101,44 @@ private function processInternalMessages(OpenDialogMessages $messageWrapper)
             }
         }
     }
+
+    /**
+     * @param UtteranceInterface $utterance
```
stu-greenshoots: I would say lets add a description to the function phpdoc and not have comments through the code flow

PR #271, opened 2019-11-20
Virtual intent data structures

elliotmassen:
This PR adds various data structures & methods that build the foundations of our virtual intent support. This PR **does not** implement the virtual intent behaviour - that will be a subsequent PR. The work in this PR allows a conversation designer to specify that an outgoing intent has a virtual intent and enables the in-memory and persisted graph to store that virtual intent.

This PR includes:
- A new `VirtualIntent` Node class.
- A new `EIModelVirtualIntent` EI model class.
- Adding a `VirtualIntent` type to DGraph.
- Updates to the `ConversationBuilder`.
- Tests that ensure that virtual intents are properly detected, stored & queried.


tests/TestCase.php
```diff
@@ -602,4 +603,41 @@ public function getMarkupForManyIntentConversation(): string
...
+  id: with_virtual_intent
+  scenes:
+    opening_scene:
+      intents:
+          - u:
+              i: intent.app.welcome
+          - b:
+              i: intent.app.welcomeResponse
+              u_virtual:
+                i: intent.app.continue
```
istos: We need another test with multiple scenes and the virtual intent taking from one scene to the other.

elliotmassen: There will be a test of this in the subsequent PR - I didn't include it here to ensure this PR just related to the data structures themselves and not the implementation.

elliotmassen:
> Looking good - I think we are not catering for a virtual intent directive defining a scene as well though.

@istos: I remember a brief conversation we had with Stu where we decided that the `scene` would be defined on the parent intent, not the virtual intent.

tests/TestCase.php
stu-greenshoots: Yes, I agree with this - so that the conversation is structured in the same way whether or not the virtual intent is present.

stu-greenshoots:
> > Looking good - I think we are not catering for a virtual intent directive defining a scene as well though.
> 
> @istos: I remember a brief conversation we had with Stu where we decided that the `scene` would be defined on the parent intent, not the virtual intent.

Yes, I agree with this - so that the conversation is structured in the same way whether or not the virtual intent is present.

PR #272, opened 2019-11-20
Feature/accessor implimentation

madan95:


src/Attribute/Composite/AbstractCompositeAttribute.php
```diff
@@ -40,17 +51,48 @@ public function setValue($value): void
...
+                    $attributes = array_reduce($attributes, function ($carry, $attribute) use ($search) {
+                        if ($attribute->getId() === $search) {
+                            $carry = $attribute;
+                        }
+                        return $carry;
+                    });
+                } elseif ($attributes instanceof AbstractAttribute) {
+                    $attributes = $attributes->getValue([$search]);
+                } else {
+                    Log::warning("Couldn't recognize attribute type in AbstractCompositeAttribute.");
```
stu-greenshoots: Should this return a value at this point?

src/Attribute/Composite/AbstractCompositeAttribute.php
```diff
@@ -40,17 +51,48 @@ public function setValue($value): void
...
+                $search = $index[$count];
+
+                if (is_array($attributes)) {
+                    $attributes = array_reduce($attributes, function ($carry, $attribute) use ($search) {
+                        if ($attribute->getId() === $search) {
+                            $carry = $attribute;
+                        }
+                        return $carry;
+                    });
+                } elseif ($attributes instanceof AbstractAttribute) {
```
stu-greenshoots: Can this check for AttributeInterface

src/Attribute/Composite/AbstractCompositeAttribute.php
```diff
@@ -62,4 +104,12 @@ public function toString(): string
     {
         return $this->attributeCollection->toString();
     }
+
+    /**
+     * @return string
+     */
+    public function getSerialized(): string
```
stu-greenshoots: Do we need this is if just does the same as the parent class?

stu-greenshoots:
@madan95 - I have picked up some of my issues on this branch and will now merge into develop (and your other open branches)

PR #275, opened 2019-11-21
Adding support for virtual intents

elliotmassen:
This PR adds support for virtual incoming intents and is dependent on work in #268 & #271. 

## Example

Virtual intents allow a conversation designer to specify that a given intent will simulate another intent. In the example below, a user would begin the conversation by saying the `intent.app.welcome` intent. After this they would, as usual, receive the `intent.app.welcomeResponse`, but also the `intent.app.endResponse`. This second intent is triggered by the `u_virtual` directive on the `intent.app.welcomeResponse` intent. The directive specifies the id of the intent to simulate, which matches the id of the succeeding incoming intent. From a chatbot user's perspective they receive two sets of messages, one from each intent.

```yaml
conversation:
  id: with_virtual_intent
  scenes:
    opening_scene:
      intents:
          - u:
              i: intent.app.welcome
          - b:
              i: intent.app.welcomeResponse
              u_virtual:
                i: intent.app.continue
          - u:
              i: intent.app.continue
          - b:
              i: intent.app.endResponse
              completes: true
```

### ConversationEngine
The `getNextIntents` method has been changed to allow for it to be called recursively. The changes are premised by a boolean parameter (`$isVirtual`), which is false by default. When a recursive call is made, the value is passed as true. This allows us to know whether we are dealing with the initial call or a recursive call to the method.

When an outgoing intent that has a virtual incoming intent is triggered, the engine checks the following incoming intents and those that don't match the id of the virtual intent will be filtered out. The remaining will be filtered by their conditions. If no intents survive the filtering, then (currently) this PR returns a NoMatchResponse along with the originally triggered outgoing intent. It doesn't seem like this would ever happen unless the conversation was incorrectly design. It may be worth considering adding validation for this in the conversation editor, as well as considering whether returning a NoMatch in this way is the desired behaviour.

## Before merging
- [x] Update the base branch to the 'develop' branch
- [x] Resolve merge conflicts

### DIFF
--- a/src/ConversationEngine/ConversationEngine.php
+++ b/src/ConversationEngine/ConversationEngine.php
@@ -18,11 +18,13 @@ use OpenDialogAi\ConversationEngine\ConversationStore\ConversationStoreInterface
 use OpenDialogAi\ConversationEngine\ConversationStore\EIModelCreatorException;
 use OpenDialogAi\ConversationEngine\ConversationStore\EIModels\EIModelCondition;
 use OpenDialogAi\ConversationEngine\ConversationStore\EIModels\EIModelIntent;
+use OpenDialogAi\ConversationEngine\Exceptions\NoMatchingIntentsException;
 use OpenDialogAi\Core\Attribute\AttributeInterface;
 use OpenDialogAi\Core\Conversation\Condition;
 use OpenDialogAi\Core\Conversation\Conversation;
 use OpenDialogAi\Core\Conversation\Intent;
 use OpenDialogAi\Core\Conversation\Scene;
+use OpenDialogAi\Core\Conversation\VirtualIntent;
 use OpenDialogAi\Core\Graph\Node\NodeDoesNotExistException;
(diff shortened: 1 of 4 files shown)


src/ConversationEngine/ConversationEngine.php
```diff
@@ -89,34 +91,33 @@ public function setActionEngine(ActionEngineInterface $actionEngine): void
...
+     * @throws CurrentIntentNotSetException
+     * @throws EIModelCreatorException
      * @throws FieldNotSupported
      * @throws GuzzleException
      * @throws NodeDoesNotExistException
-     * @throws EIModelCreatorException
-     * @throws CurrentIntentNotSetException
      */
-    public function getNextIntents(UserContext $userContext, UtteranceInterface $utterance): array
+    public function getNextIntents(UserContext $userContext, UtteranceInterface $utterance, bool $isVirtual = false): array
```
stu-greenshoots: Could we put in a summary of what the method does in the phpdoc.

PR #278, opened 2019-11-26
Using loose comparison in AbstractAttributeCollection

madan95:


stu-greenshoots:
Good catch, thanks @madan95 .

PR #283, opened 2019-11-28
Adds Microsoft language support via a NlpEngine

duncanheron:
Creates a new NlpEngine to hide the complexity of differing NLP service calls


src/NlpEngine/Client/MsClient.php
```diff
@@ -0,0 +1,23 @@
...
+use GuzzleHttp\Client;
+
+class MsClient extends Client
+{
+    /**
+     * MsClient constructor.
+     *
+     * @param array $config
+     */
+    public function __construct(array $config = [])
```
stu-greenshoots: I would say this is better to be bound in the service provider instead of creating an extension of the Guzzle Client.

So doing something like
```$this->singleton('MsClient, fuunction () {
$client  = new Guzzle/Client();
// set up with values from config

return $client;
});

Or have a look at contextual binding to supply a custom Guzzle Client to the MS Service:
https://laravel.com/docs/5.8/container#contextual-binding.

File: src/NlpEngine/Service/MsNlpService.php [hunk 2]
```diff
@@ -0,0 +1,56 @@
...
+        ];
+
+        $this->client->post(
+            '/languages',
+            [
+                'form_params' => $body
+            ]
+        );
+
+        $language = new NlpLanguage();
```
stu-greenshoots: Should be using the response here :)

src/NlpEngine/Service/MsNlpService.php
```diff
@@ -0,0 +1,56 @@
...
+
+    public function __construct(string $string, MsClient $client)
+    {
+        $this->client = $client;
+        $this->string = $string;
+    }
+
+    public function getLanguage(): NLPLanguage
+    {
+        $body = [
```
stu-greenshoots: I'm almost tempted to say that it might be nice to extract the API body set up and calls to a separate Client class similar to what we do with LUIS client etc. Then the service class can make a call to 
```
$this->client->getLangauge($this->string);
```

And get the body back. Would allow us to centralise error handling easily as well

src/NlpEngine/Service/MsNlpService.php
```diff
@@ -0,0 +1,56 @@
...
+ */
+class MsNlpService implements NlpServiceInterface
+{
+    /** @var \OpenDialogAi\Core\NlpEngine\Client\MsClient  */
+    private $client;
+
+    /** @var string  */
+    private $string;
+
+    const LANGUAGE_DEFAULT = 'GB';
```
stu-greenshoots: This would be nice pulled up to the config level (and then passed in from the service provider so that the class doesn't need to make calls out to `config()`

tests/TestCase.php
```diff
@@ -598,4 +600,12 @@ public function getMarkupForManyIntentConversation(): string
 EOT;
         return $conversationMarkup;
     }
+
+    /**
+     * @return string
+     */
+    protected function getTestStringForNlp(): string
```
stu-greenshoots: Can we scope this to the NLP tests? A Trait or something like that would work

stu-greenshoots: Or even just having it repeated in the test classes works for me to be honest

src/NlpEngine/Service/MsNlpService.php
stu-greenshoots: Let me know what you think about this. If we go with this, then the binding I put in the comment above is irrelevant

PR #289, opened 2019-12-04
Adding support for DGraph facets

elliotmassen:
This PR is the first of four that will enable the 'repeating' intent feature. This PR specifically adds support for [DGraph facets](https://docs.dgraph.io/query-language/#facets-edge-attributes) and adds one to the `followed_by` edge that connects intents in a user conversation.

It's worth mentioning that facet support has only been added to what OpenDialog would call 'relationships' between nodes and is not available for node 'attributes'. Although DGraph itself supports facets on any edge, OpenDialog does not use `Edge`'s to connect node attributes in-memory.

### DGraph integration
Our DGraph integration has been updated to support facets. This means that we are now able to specify that we would like to query facets in our DGraph queries. A new constant has been added to specify this: `DGraphQuery::WITH_FACETS`. This can be used alone to get all facets or be used as an array key to details specific facets to query.

### New class 'ModelFacets'
A new class has been added called `ModelFacets`. The intention is that this will be a similar class to `Model` in that we can use it to centralise facet names used in the graph. There is also a static method for correctly formatting the full facet names expected in DGraph query responses.


src/Conversation/Intent.php
```diff
@@ -476,4 +476,30 @@ public function getVirtualIntent(): ?VirtualIntent
...
+    public function getFollowedBy(): Intent
+    {
+        return $this->getNodesConnectedByOutgoingRelationship(Model::FOLLOWED_BY)->first()->value;
+    }
+
+    /**
+     * @param Intent $toIntent
+     * @param $createdAt
+     */
+    public function setFollowedBy(Intent $toIntent, $createdAt)
```
stu-greenshoots: The create outgoing edge method is a bit confusing, as it looks like the newly created DirectedEdge is never used or persisted. Discussed with Elliot and I can see what's going on now, but I will raise an issue to clarify this process a bit

src/ConversationEngine/ConversationStore/EIModels/EIModelIntent.php
```diff
@@ -29,6 +30,10 @@ class EIModelIntent extends EIModelWithConditions
 
     private $nextScene;
 
+    private $followedBy;
```
stu-greenshoots: Can these be type hinted

src/Graph/DGraph/DGraphQuery.php
```diff
@@ -184,9 +185,28 @@ public function decodeQueryGraph($queryGraph)
...
-                $result .= $item . ' ';
+                if ($item === DGraphQuery::WITH_FACETS) {
+                    $result .= $key . ' @facets';
+                } else {
+                    $result .= $item . ' ';
+                }
             } else {
                 $result .= $key;
+
+                if (array_key_exists(DGraphQuery::WITH_FACETS, $item) || in_array(DGraphQuery::WITH_FACETS, $item)) {
```
stu-greenshoots: I think this would be easier to read with just an if / else if or maybe moving the logic of the if statement into a method that returns a boolean? eg
`function itemHasFacets($item): bool`

src/ContextEngine/tests/UserServiceTest.php
```diff
@@ -45,10 +48,7 @@ public function setUp(): void
         $this->client->dropSchema();
         $this->client->initSchema();
 
-        for ($i = 1; $i <= 4; $i++) {
-            $conversationId = 'conversation' . $i;
-            $this->activateConversation($this->$conversationId());
-        }
+        $this->activateConversation($this->conversation1());
```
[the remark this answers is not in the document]
elliotmassen: That's right - none of the tests made use of the other conversations at all.

src/Graph/DGraph/DGraphQuery.php
elliotmassen: That's a good idea - I've now updated this, let me know what you think!

PR #291, opened 2019-12-05
Add Util methods to encode/decode

madan95:
**Description**
- Changes: Utils method for encoding/decoding in Attributes
- Reason: DRY, Single source of truth for encoding/decoding format in Attributes, reduce noise in codebase


stu-greenshoots:
Good idea, I like it. Reuse for the win

PR #293, opened 2019-12-06
Allow to add html tags in the button text

nicolocaruso85:


src/ResponseEngine/Formatters/Webchat/WebchatMessageFormatter.php
```diff
@@ -421,9 +421,28 @@ private function formatButtonTemplate(SimpleXMLElement $item): array
...
+                foreach ($dom->childNodes as $node) {
+                    foreach ($node->childNodes as $item) {
+                        if ($item->nodeType === XML_TEXT_NODE) {
+                            if (!empty(trim($item->textContent))) {
+                                $buttonText .= ' ' . trim($item->textContent);
+                            }
+                        } elseif ($item->nodeType === XML_ELEMENT_NODE) {
+                            if (!empty(trim($item->textContent))) {
+                                $buttonText .= ' ';
+                                $buttonText .= sprintf('<%s>%s</%s>', $item->nodeName, trim($item->textContent), $item->nodeName);
```
elliotmassen: Please restrict this to only allow bold, italic and underline tags. In OD message format these will be represented by `<b></b>`, `<i></i>` & `<u></u>` respectively, and should be converted to `<strong></strong>`, `<em></em>` & `<u></u>` respectively in the webchat formatter.

src/ResponseEngine/Formatters/Webchat/WebchatMessageFormatter.php
```diff
@@ -421,9 +421,45 @@ private function formatButtonTemplate(SimpleXMLElement $item): array
...
+                                    $buttonText .= sprintf(
+                                        '<strong>%s</strong>',
+                                        trim($item->textContent)
+                                    );
+                                } elseif ($item->nodeName === 'i') {
+                                    $buttonText .= sprintf(
+                                        '<em>%s</em>',
+                                        trim($item->textContent)
+                                    );
+                                } else {
```
elliotmassen: This `else` should be an `elseif` for checking the `<u>` tag. Without it, any tag could be used - the tags should be limited.

src/ResponseEngine/Rules/MessageXML.php
```diff
@@ -63,16 +63,23 @@ public function passes($attribute, $value)
                             return false;
                         }
                         foreach ($item->button as $button) {
+                            $buttonXml = $button->text->asXml();
+                            $buttonText = trim(substr(substr($buttonXml, 0, - 7), 6));
```
elliotmassen: I think this should be a bit more robust. Perhaps in the future we may have attributes on the `text` node for example. It looks like the `__toString()` method may be useful at giving the internal contents of a node without the enclosing tags.

elliotmassen:
@nicolocaruso85 I took a look at the most recent changes, but it seems that the message isn't considered valid for some reason. I've added to the test to ensure that the validation is being tested also.

PR #294, opened 2019-12-06
Repeating intent data structures

elliotmassen:
This PR adds various methods and properties to enable a conversation designer to specify that an intent is a repeating intent. It is related to work in #289 as this PR is one in a set of PR's that will enable repeating intent functionality.


src/Conversation/Intent.php
```diff
@@ -41,6 +42,7 @@ public function __construct($id, $completes = false)
     {
         parent::__construct($id);
         $this->addAttribute(new StringAttribute(Model::EI_TYPE, Model::INTENT));
+        $this->addAttribute(new StringAttribute(Model::REPEATING, false));
```
stu-greenshoots: Seems like this should be a BooleanAttribute

stu-greenshoots: I pushed in an update that uses a boolean

PR #297, opened 2019-12-10
Feature/bubble field not supported exception

madan95:
**Description**
We are starting to see pattern/boiler plate where we need to catch FieldNotSupported Exception on most of the interpreter. 

**Solution**
Let the exception bubble up to the caller (ConversationEngine) & handle/log exception there.

### DIFF
--- a/src/ConversationEngine/ConversationEngine.php
+++ b/src/ConversationEngine/ConversationEngine.php
@@ -9,9 +9,9 @@ use Illuminate\Support\Facades\Log;
 use OpenDialogAi\ActionEngine\Actions\ActionResult;
 use OpenDialogAi\ActionEngine\Exceptions\ActionNotAvailableException;
 use OpenDialogAi\ActionEngine\Service\ActionEngineInterface;
+use OpenDialogAi\ContextEngine\ContextManager\ContextInterface;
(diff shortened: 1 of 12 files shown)


src/ConversationEngine/ConversationEngine.php
```diff
@@ -411,9 +429,17 @@ private function matchOpeningIntents(
...
-                $intentsFromInterpreter = $this->interpreterService
-                    ->getInterpreter($validIntent->getInterpreterId())
-                    ->interpret($utterance);
+                try {
+                    $intentsFromInterpreter = $this->interpreterService
+                        ->getInterpreter($validIntent->getInterpreterId())
+                        ->interpret($utterance);
+                } catch (FieldNotSupported $e) {
+                    Log::warning(sprintf(
+                        'Trying to use %s interpreter to interpret an utterance/field that is not supported.',
```
stu-greenshoots: I think this logic is repeating too much. We could centralise this logic into a new `interpret` method in the interpreter service that takes the name of the interpreter to run and runs it's `interpret` method.

stu-greenshoots:
@nicolocaruso85 - thanks for updating this to use the interpreter service. Can you also pull in the try/catch logic for FieldNotSupported so that it is not repeated?

PR #299, opened 2019-12-13
Hide new message property “hide_avatar”

nicolocaruso85:


src/ResponseEngine/Message/Webchat/WebchatMessage.php
```diff
@@ -136,6 +138,26 @@ public function setHidetime($hidetime)
         return $this;
     }
 
+    /**
+     * @return bool
+     */
+    public function getHideavatar()
+    {
+        return $this->hideavatar;
+    }
```
elliotmassen: Please update the `hideavatar` property & method names to use the usually lower-camel-case style.

Issue #282, opened 2019-11-28
Actions should not run if they have not been passed their required attributes


stu-greenshoots:
In the action engine, we make a check as to whether the required attributes required for the action have been defined in the `input attributes` on the conversation here:

https://github.com/opendialogai/core/blob/develop/src/ActionEngine/Service/ActionEngine.php#L79

But we do not check whether the attributes stated actually exist or not. This is compounded by the issue that the context service returns an empty string attribute in the case that it cannot find the specified attribute:

https://github.com/opendialogai/core/blob/develop/src/ContextEngine/ContextManager/ContextService.php#L187

## What we want

Firstly, the Context Service should stop returning an empty attribute in the case that it can't find one and throw the `AttributeDoesNotExistException`

This would mean that it is up to the called of the context service to catch exceptions and deal with them. We could think about putting in a `getAttributeOrNull` type method that would handle the exception and return an empty String if the attribute is not found

Then the action engine can check for the Exception when building up the `ActionInput`, and if it is thrown, a warning should be logged and the action should not be run.

PR #301, opened 2019-12-16
Feature/action input validation

stu-greenshoots:
This PR goes a bit further than just validating action input. It:

- Validates that the action input contains all the required parameters AFTER resolving the attributes (ie if an attribute doesn't exist, the action is not run)
- Stops the context service returning an empty string attribute when the specified attribute does not exist 
- Adds a method for unsuccessful action results (which is used to return one when the action isn't run.
- Improves logging slightly in the ResponseEngineService
- Stops the operation service trying to run an operation that isn't provided the required attribute

Fixes https://github.com/opendialogai/core/issues/282


src/OperationEngine/Service/OperationService.php
```diff
@@ -116,4 +102,60 @@ public function checkCondition(Condition $condition)
 
         return $operation->execute();
     }
+
+    /**
+     * @param Condition $condition
+     * @param ParsedAttributeName $parsedAttributeName
+     * @return AttributeInterface
+     */
+    private function getActualAttribute(Condition $condition, ParsedAttributeName $parsedAttributeName): AttributeInterface
```
elliotmassen: I'm wondering if this method could be named differently. Perhaps just `getParsedAttribute` or `getAttribute`?

Issue #302, opened 2019-12-19
getAttributeValue doesn't gracefully handle empty attributes


duncanheron:
Calling `ContextService::getUserContext()->getAttributeValue('nonexistentvalue')` when an attribute doesnt exists currently throws a fatal error

stu-greenshoots:
This issue is still open as we are seeing issues on the user context with getting attribute values directly always returning null

stu-greenshoots:
https://github.com/opendialogai/core/issues/318

PR #308, opened 2020-01-03
Return an empty value when call getAttributeValue if the attribute doesnt exists

nicolocaruso85:


nicolocaruso85:
fixes https://github.com/opendialogai/core/issues/302

Issue #274, opened 2019-11-21
Message conditions fail if they cannot be read


stu-greenshoots:
If conditions on a message are not valid (and it is very easy to not structure a condition correctly as the names of values depend on the implementation of the operation!) the conversation breaks and the user gets an error message. 

We should handle this in the response builder, and if a condition cannot be read, it should return a false and log a warning

PR #309, opened 2020-01-03
Before execute an operation check if all required parameters are there

nicolocaruso85:
fixes https://github.com/opendialogai/core/issues/274


stu-greenshoots:
Thanks for the update, merging in

PR #317, opened 2020-01-09
Add tests for Import and Export a Conversation

nicolocaruso85:


src/Console/Commands/tests/ImportExportConversationTest.php
```diff
@@ -0,0 +1,31 @@
...
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        $this->initDDgraph();
+    }
+
+    public function testImportConversation()
+    {
+        $exitCode = Artisan::call('conversation:import ' . dirname(__FILE__) . '/example_conversation --activate --yes');
```
stu-greenshoots: Can you add an assertion to check that the conversation is actually loaded into the database and DGraph

Issue #318, opened 2020-01-09
getAttributeValue on user context always returning a null


stu-greenshoots:
I have created a new branch with a failing test to highlight the issue: bug/get-attribute-value

Other contexts work as expected, the user context always returns a null

PR #321, opened 2020-01-10
Bug/get attribute value

nicolocaruso85:
fixes https://github.com/opendialogai/core/issues/318

### DIFF
--- a/src/ContextEngine/Contexts/User/UserContext.php
+++ b/src/ContextEngine/Contexts/User/UserContext.php
@@ -74,12 +74,16 @@ class UserContext extends AbstractContext
     /**
      * @param string $attributeName
      * @return mixed
-     * @throws AttributeDoesNotExistException
      */
     public function getAttributeValue(string $attributeName)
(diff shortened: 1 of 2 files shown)


src/ContextEngine/Contexts/User/UserContext.php
```diff
@@ -71,6 +71,17 @@ public function getAttribute(string $attributeName): AttributeInterface
...
 
+    /**
+     * @param string $attributeName
+     * @return mixed
+     * @throws AttributeDoesNotExistException
+     */
+    public function getAttributeValue(string $attributeName)
+    {
+        $attribute = $this->getAttribute($attributeName);
+        return $attribute->getValue();
```
elliotmassen: For consistency, I think it would be worth having this method be similar to `HasAttributesTrait.getAttributeValue`, especially that it returns null. The change would be that rather than calling `$this->hasAttribute` it calls `$this->hasUserAttribute`.

PR #322, opened 2020-01-10
NLP Engine updates

elliotmassen:
This PR updates the NLP Engine to make it possible to register custom NLP providers in an OpenDialog application. These changes bring the engine more inline with the structure that we use in the Interpreter Engine. As such there are a number of breaking changes that we need to be aware of:

1. The NLP engine config is now comprised of two files `opendialog-nlpengine.php` and `opendialog-nlpengine-custom.php`. These work in the same way as in other engines: the standard config is for core level configuration, while the custom config is published to the OpenDialog application so that a chatbot developer can add additional configuration.
2. This PR introduces a `NlpService` which can be used to access the registered providers. Previously the NLP providers were called NLP services, but as we now have an `NlpService` these have been renamed to avoid confusion. There may be a better name than providers.
3. Prior to this PR, NLP providers were instantiated when they were needed. We can now retrieve provider objects from `NlpService.getProvider`. Due to this, we can no longer pass in a document to a provider's constructor. Each NLP method (`getSentiment`, `getEntities`, etc) will now take the document as a parameter.

## Outstanding actions
- [x] Decide whether we are happy with the name of NLP "providers".


src/NlpEngine/MicrosoftRepository/MsClient.php
```diff
@@ -1,21 +1,30 @@
...
 {
-    /** @var \GuzzleHttp\Client  */
+    /** @var Client */
     private $client;
 
-    public function __construct()
+    /**
+     * @param Client $client
+     */
+    public function setClient(Client $client): void
```
duncanheron: Not sure if its just personal preferrence here but should __construct come first?

src/NlpEngine/tests/MsNlpProviderTest.php
```diff
@@ -2,43 +2,59 @@
...
     }
 
     public function testItGetsLanguageFromMs()
     {
         $this->clientMock->shouldReceive('getLanguage')->once()->andReturn($this->getTestResponse());
-        $language = $this->msNlpService->getLanguage();
+
+        try {
+            $language = $this->msNlpProvider->getLanguage($this->getTestStringForNlp());
+        } catch (NlpProviderMethodNotSupportedException $e) {
```
duncanheron: should we have a test that these custom exceptions are thrown?

PR #324, opened 2020-01-15
Adding confidence to AbstractNLUInterpreter logging

elliotmassen:
This PR adds the confidence of an interpreted Rasa or LUIS intent to the logged message. This will give us better insight into our NLU interpreters when an intent isn't interpreted as expected - usually it is because the confidence did not meet the desired threshold.


stu-greenshoots:
Works as expected, so merging this in. It might be an idea in the future to move this logging logic to the interpreter service so that all intents get logged

PR #325, opened 2020-01-17
Improve Conversation Yaml validation

nicolocaruso85:


nicolocaruso85:
fixes https://github.com/opendialogai/core/issues/326

stu-greenshoots:
I think we need to review how we are doing validation - what we have at the moment is really brittle and very much tied to our current YAML structure with no easy wat to update it. I will close this PR, but leave the branch intact so we can see the work for reference

PR #329, opened 2020-01-27
Export only required conversation fields

nicolocaruso85:


elliotmassen:
@stu-greenshoots I'm wondering whether we should have `Conversation`, `OutgoingIntent` & `MessageTemplate` models define some sort of public `exportableFields` array, which the export command can use - as opposed to hard coding the fields in the command and spreading out the control away from the models. What do you think?

PR #338, opened 2020-02-28
Add new message type “Hand To Human message”

nicolocaruso85:


src/ResponseEngine/tests/ResponseEngineTest.php
```diff
@@ -243,6 +244,47 @@ public function testWebChatMessage()
         $this->assertInstanceOf('OpenDialogAi\ResponseEngine\Message\OpenDialogMessage', $messageWrapper->getMessages()[0]);
     }
 
+    public function testWebChatHandToHumanMessage()
```
elliotmassen: Please could you add assertions for the message content here. It looks like they will be a problem in the new MessageBuilder class' `getMarkup` method.

PR #340, opened 2020-03-02
Update user service to add a new method that takes a user id and returns whether the user is a new, returning or ongoing user

nicolocaruso85:


stu-greenshoots:
@nicolocaruso85 please always add tests for functions like this. I have had to add the tests myself here as we need the functionality ASAP

PR #364, opened 2020-05-14
Add Dialogflow interpreter

nicolocaruso85:


src/InterpreterEngine/Interpreters/DialogflowInterpreter.php
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        return [new NoMatchIntent()];
+    }
+
+    /**
+     * @param $text
+     * @return string
+     */
+    private function formatLinks($text)
```
istos, line 86: This should be in a utility class in the InterpreterEngine message space - other interpreters will almost certainly need it

src/InterpreterEngine/Interpreters/DialogflowInterpreter.php
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /**
+     * @param UtteranceInterface $utterance
+     * @param DialogflowClient $client
+     * @return array
+     * @throws \OpenDialogAi\Core\Utterances\Exceptions\FieldNotSupported
+     * @throws \OpenDialogAi\InterpreterEngine\Exceptions\InterpreterNameNotSetException
+     */
+    public function interpretWithClient(UtteranceInterface $utterance, DialogflowClient $client): array
```
istos, line 38: Although the Dialogflow client allows us to choose a project id to interpret with we have no way in the interpreter to set that, we should provide a `setDialogflowProjectId` and `getDialogflowProjectId` in the interpreter (even though we still don't have a way to pass that through the

PR #382, opened 2020-08-06
Allow u_virtual to work on a completing intent

stu-greenshoots:
updates the flow of the conversation engine to update the user's current position after each intent is resolved. This allows an intent with a `u_virtual` directive to complete a conversation and move the user into a new conversation with a matching opening intent


stu-greenshoots:
> I haven't had a proper chance to look at this, but at a glance this looks good and makes sense to me 👍 I was wondering if there are any repercussions for moving the functionality that determines `$isRepeating` & `$precedingIntent` but after having a think their move doesn't seem to break anything as far as I can see.

Yeah, that's in my mind as well, but seems good. With this change, we still always perform those actions but instead do it after each bot intent that's resolved rather than at the end. I could see this leading to an accidental loop if repeating and u_virtual are used together in strange ways.

I'll keep an eye on it through the MMM development

PR #384, opened 2020-08-10
adds submit_text and callback fields to the autocomplete message

stu-greenshoots:
updates the original PR to add fields for callback and submit text

### DIFF
--- a/src/MessageBuilder/Message/AutoCompleteMessage.php
+++ b/src/MessageBuilder/Message/AutoCompleteMessage.php
@@ -21,18 +21,18 @@ class AutoCompleteMessage
      * @param $title
      * @param $endpointUrl
      * @param $queryParamName
-     * @param array $endpointParams
      * @param $callback
      * @param $submitText
+     * @param array $endpointParams
      */
     public function __construct($title, $endpointUrl, $queryParamName, $callback, $submitText, $endpointParams = [])
     {
         $this->title = $title;
         $this->endpointUrl =$endpointUrl;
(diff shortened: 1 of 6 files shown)


src/MessageBuilder/Message/AutoCompleteMessage.php
```diff
@@ -12,25 +12,36 @@ class AutoCompleteMessage
...
+
+    public $submitText;
+
     /**
      * AutoCompleteMessage constructor.
      * @param $title
      * @param $endpointUrl
-     * @param $endpointParams
+     * @param $queryParamName
+     * @param array $endpointParams
```
patshone-gsl, line 24: I'm assuming that the order of the params in the doc here should match the constructor - ie the endpointParams at the end

src/MessageBuilder/Message/AutoCompleteMessage.php
```diff
@@ -12,25 +12,36 @@ class AutoCompleteMessage
...
+     * @param $callback
+     * @param $submitText
      */
-    public function __construct($title, $endpointUrl, $queryParamName, $endpointParams = [])
+    public function __construct($title, $endpointUrl, $queryParamName, $callback, $submitText, $endpointParams = [])
     {
         $this->title = $title;
         $this->endpointUrl =$endpointUrl;
         $this->queryParamName= $queryParamName;
         $this->endpointParams = $endpointParams;
```
patshone-gsl, line 33: And similarly for the ordering here - setting the endpointParams at the end

PR #411, opened 2020-09-18
ensures the microtime will always return decimal values

stu-greenshoots:
We have seen an issue in the Message class where the error `Call to a member function format() on bool` is seen in the logs and the user receives an error in the chatbot.

This fix updates the part of the core where `microtime` is used to make sure that the value is number formatted to include trailing 0s even when it hits an exact second.


istos:
@stu-greenshoots what is sensitive about microtime in this way?

PR #415, opened 2020-11-06
Composer 2 PSR-4 Compliance

elliotmassen:
This PR updates various namespaces to be compliant with PSR-4, which Composer 2 is now requiring. The changes are limited to test classes, however is seems that the Code Fixer has also made formatting changes to a few other classes (import reordering, `else if` to `elseif` conversions, spacing, etc).


stu-greenshoots (approved):
Thanks for this, has been needed for a while :)

PR #417, opened 2020-11-18
Allow button message text element to be empty

elliotmassen:
This PR allows for button message text to be `null` which is already accounted for by [Webchat](https://github.com/opendialogai/webchat). This means that you can have buttons without needing preceding text.

### DIFF
--- a/src/ResponseEngine/tests/ResponseEngineWebchatMessageFormatterTest.php
+++ b/src/ResponseEngine/tests/ResponseEngineWebchatMessageFormatterTest.php
@@ -200,6 +200,7 @@ EOT;
 
         $this->assertEquals(true, $message->getData()['clear_after_interaction']);
         $this->assertEquals(true, $message->getData()['disable_text']);
+        $this->assertEquals('test', $message->getData()['text']);
         $this->assertEquals($expectedOutput, $message->getButtonsArray());
 
         $markup = <<<EOT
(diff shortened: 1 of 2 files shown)


src/ResponseEngine/tests/ResponseEngineWebchatMessageFormatterTest.php
```diff
@@ -286,6 +287,30 @@ public function testButtonMessage()
 
         $this->assertEquals(false, $message->getData()['clear_after_interaction']);
         $this->assertEquals(false, $message->getData()['disable_text']);
+        $this->assertEquals('test', $message->getData()['text']);
+        $this->assertEquals($expectedOutput, $message->getButtonsArray());
+
+        // phpcs:ignore
+        $markup = '<message><button-message><text></text><button type="no-button"><text>No</text><callback>callback_no</callback><value>false</value></button></button-message></message>';
```
stu-greenshoots, line 294: Do we still have to out in the `<text></text>`? 

Is that because of validation?

stu-greenshoots (changes requested):

src/ResponseEngine/tests/ResponseEngineWebchatMessageFormatterTest.php
elliotmassen: I think the editor would require this, yes.

PR #418, opened 2020-12-11
Graph schema configuration

elliotmassen:
This PR allows additional graph schema to be defined at the application level. A new config file (`opendialog/graph.php`) will now be published to the application which allows for this additional schema to be defined. There is currently now validation of this additional schema, so care should be taken to ensure that existing predicate and type names are not duplicated as this will cause unexpected issues in the application.


patshone-gsl (approved):
LGTM

PR #419, opened 2021-01-05
Feature/default string attribute

stu-greenshoots:
Updates AttributeResolver::getAttributefor() method to default to a string type attribute if the given attribute ID has not been registered where it used to throw an AttributeIsNotSupported exception.

This means that if chatbot builders can use attributes in the conversation and message flow without having first given them type.

This could cause something unexpected if the attribute is not a string, and is used in a condition.

All sections of the code that expected the AttributeIsNotSupported exception have been updated to remove the try/catch blocks.


stu-greenshoots:
After the first commit, there were some failing tests that I wouldn't have expected to fail (ActionEngineTest, ContextEngineServiceTest, ResponseEngineTest).

I have fixed the tests, but will think a bit more on what has changed to cause them to fail

istos (approved):

Issue #420, opened 2021-01-12
Failed DGraph mutation causes an ErrorException


elliotmassen:
If a DGraph mutation is not successful and therefore doesn't have a `data` property then `DGraphMutationResponse.data` is never set. However, in the `isSuccessful` method the following condition attempts to index it as an array. This causes an `ErrorException`: "Trying to access array offset on value of type null".

https://github.com/opendialogai/core/blob/3290352ce31ee862b885f424a9880d429678d3ac/src/Graph/DGraph/DGraphMutationResponse.php#L50

PR #427, opened 2021-01-20
Null check on DGraph Response

stu-greenshoots:
This PR fixes issue https://github.com/opendialogai/core/issues/420 by first checking if the data element has been set before checking the `code` within it.

It seems like this issue will only affect projects from PHP 7.4 onwards (previously, PHP would return a null in this case which would fail the check as expected in the code flow)

We have seen an uncaught exception thrown from this code in some projects, and in theory, this fix should allow the code to run as expected avoiding the issues we've seen


elliotmassen (approved):
Perfect 👍

PR #428, opened 2021-01-26
OPNDLG-330 Dynamic Attributes

bendeeson:
Adds the ability to define attributes dynamically in the database instead of only in config files.

Adds a new Laravel Model 'Dynamic Attribute' which stores an attribute id (unique snake_case string) and attribute type (attribute.core.<name>) in the database. Includes two regex patterns to check the format of the attribute id and type.

Intended for use in conjuction with this PR for opendialog: https://github.com/opendialogai/opendialog/pull/233

### DIFF
--- a/database/migrations/2021_01_22_160714_create_dynamic_attributes_table.php
+++ b/database/migrations/2021_01_22_160714_create_dynamic_attributes_table.php
@@ -15,6 +15,7 @@ class CreateDynamicAttributesTable extends Migration
     {
         Schema::create('dynamic_attributes', function (Blueprint $table) {
             $table->bigIncrements('id');
+            $table->timestamps();
             $table->string('attribute_id')->unique();
             $table->string('attribute_type');
         });
--- a/database/migrations/2021_01_22_160714_create_dynamic_attributes_table.php
+++ b/database/migrations/2021_01_22_160714_create_dynamic_attributes_table.php
@@ -14,9 +14,9 @@ class CreateDynamicAttributesTable extends Migration
     public function up()
     {
         Schema::create('dynamic_attributes', function (Blueprint $table) {
-            $table->string('id')->unique();
-            $table->string('type');
-            $table->primary('id');
+            $table->bigIncrements('id');
+            $table->string('attribute_id')->unique();
+            $table->string('attribute_type');
         });
     }
 
--- a/database/migrations/2021_01_22_160714_create_dynamic_attributes_table.php
+++ b/database/migrations/2021_01_22_160714_create_dynamic_attributes_table.php
@@ -1,0 +1,32 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class CreateDynamicAttributesTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('dynamic_attributes', function (Blueprint $table) {
+            $table->string('id')->unique();
+            $table->string('type');
+            $table->primary('id');
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::dropIfExists('dynamic_attributes');
+    }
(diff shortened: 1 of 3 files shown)


src/DynamicAttribute.php
```diff
@@ -0,0 +1,44 @@
...
+use Illuminate\Database\Eloquent\Model;
+
+class DynamicAttribute extends Model
+{
+    static private $validIdPattern = "/^([a-z]+_)*[a-z]+$/";
+    static private $validTypePattern = "/^attribute\.[A-Za-z]*\.[A-Za-z_]*$/";
+
+    public $incrementing = false;
+
+    public $timestamps = false;
```
elliotmassen, line 14: I think it might be worth having the `created_at` & `updated_at` timestamps enabled, it may be useful for the UI in the near future.

database/migrations/2021_01_22_160714_create_dynamic_attributes_table.php
```diff
@@ -0,0 +1,32 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('dynamic_attributes', function (Blueprint $table) {
+            $table->string('id')->unique();
+            $table->string('type');
+            $table->primary('id');
```
elliotmassen, line 19: I'm thinking that it might be better to use an auto-incrementing int ID as the primary key. As we want each attribute to be granularly editable, I can imagine a scenario in which a user changed the name of an attribute and if we are using the name as the key, that may introduce issues with querying/routing. Let me know what you think - I might be overthinking it.

src/DynamicAttribute.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace OpenDialogAi\Core;
+
+use Illuminate\Database\Eloquent\Model;
+
+class DynamicAttribute extends Model
+{
+    static private $validIdPattern = "/^([a-z]+_)*[a-z]+$/";
+    static private $validTypePattern = "/^attribute\.[A-Za-z]*\.[A-Za-z_]*$/";
```
elliotmassen, line 10: Do you think it might be worth moving these patterns to the `AttributeResolver`? I'm looking at how we do it for interpreters and we have a similar `$validNamePattern` in the `InterpreterService`. Perhaps we get to the "contracts" work these patterns will even live there.

elliotmassen (changes requested):
This looks good, nice one 👍  Just a few thoughts below. Once addressed, I suppose we could either merge this without the integration with `ContextEngine` (or `AttributeEngine` if #429 is merged), or add the integration to this PR. I'm thinking it'd be better to do the integration in a new PR so that we can get this merged and keep things moving.

database/migrations/2021_01_22_160714_create_dynamic_attributes_table.php
bendeeson: Yes, I can see the value in having a stable ID for routing purposes. I'll change the `id` column to be the auto-incrementing and add a unique string `attribute_id` for the actual attribute id. The name `attribute_id` feels a bit clumsy, so let me know if have something else in mind.

src/DynamicAttribute.php
bendeeson: I this case I'd prefer to wait until theres a use case, and then alter the schema to suit. It would be especially confusing in the future if there are a `created_at` and `updated_at` columns, but the don't get used for anything.

database/migrations/2021_01_22_160714_create_dynamic_attributes_table.php
elliotmassen: Perhaps `attribute_name`?

src/DynamicAttribute.php
stu-greenshoots: I'm inclined to agree with @elliotmassen as we use these throughout our schema currently. I could see why an updated_at might need a use case to be useful (and arguably we'd need an activity log type thing for this case), but being able to display when the dynamic attribute was added on the UI seems like something we'd want

elliotmassen (approved):
This looks good to me! As discussed, I'll leave with you to handle merging/branching of this as you see fit.

elliotmassen:
Closing in favour of #433.

PR #429, opened 2021-01-26
AttributeEngine

elliotmassen:
This PR creates a `AttributeEngine` directory and namespace by splitting out attribute specific functionality and configuration from the `ContextEngine`. The `ContextEngine` now deals exclusively with contexts, and the various attribute types and `AttributeResolver` have been moved to the new `AttributeEngine`. This separation of concerns gives us a place to cater for attributes away from contexts, which is how we understand them conceptually.

As this includes many namespace changes of fundamental classes, this PR touches a lot of files - however the changes in-of-themselves are relatively minor. These changes are breaking and will require any applications (at least):

 - to update the imports of `AttributeResolver`
 - to update any imports of the core attribute classes
 - to run `php artisan vendor:publish --tag=opendialog-config` to get the new engine config
 - to move registration of custom attributes to the new config file

To ensure that you get Composer to pick up the new namespace, you'll need to run `composer dumpautoload` before `AttributeEngine` will be auto-loaded properly.

| Previous directory structure | New directory structure |
| --- | --- |
| Screenshot 2021-01-26 at 16 46 16 | Screenshot 2021-01-26 at 16 45 15 |


patshone-gsl (approved):
I think this PR will be a good test to validate the amount of work it takes to bring braking changes from core through to the opendialog application and shoebot

PR #430, opened 2021-01-28
AttributeTypeService

elliotmassen:
This PR adds an `AttributeTypeService` which explicitly registers core and custom attribute types. Previously there was no such registration and the `AttributeResolver` would only validate attribute types for each individual attribute registration. This new service lays the ground work to expose available attribute types in the upcoming reflection API and to enable validation of dynamic attributes (#428).

### DIFF
--- a/src/AttributeEngine/Tests/AttributeResolverServiceTest.php
+++ b/src/AttributeEngine/Tests/AttributeResolverServiceTest.php
@@ -41,15 +41,40 @@ class AttributeResolverServiceTest extends TestCase
         $this->assertEquals(StringAttribute::class, get_class($attribute));
     }
 
-    public function testBindingCustomAttributes()
+    public function testBindingCustomAttributesWithUnregisteredCustomType()
     {
+        // Don't register any attribute types
         $this->setConfigValue(
-            'opendialog.attribute_engine.custom_attributes',
-            ['test_attribute' => StringAttribute::class]
+            'opendialog.attribute_engine.custom_attribute_types',
+            []
         );
 
+        $this->setConfigValue(
+            'opendialog.attribute_engine.custom_attributes',
+            ['test_attribute' => ExampleCustomAttributeType::class]
+        );
+
+        // Our custom attribute type isn't register so we should fallback to a string attribute
         $attributeResolver = $this->getAttributeResolver();
         $this->assertEquals(StringAttribute::class, get_class($attributeResolver->getAttributeFor('test_attribute', null)));
+    }
+
+    public function testBindingCustomAttributesWithRegisteredCustomType()
+    {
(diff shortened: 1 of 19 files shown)


src/AttributeEngine/Tests/AttributeTypeServiceTest.php
```diff
@@ -0,0 +1,84 @@
...
+        $this->assertTrue($attributeTypeService->isAttributeTypeAvailable(ExampleCustomAttributeType::$type));
+    }
+
+    public function testRegisterCustomAttributeTypeWithUsedId()
+    {
+        $attributeTypeService = resolve(AttributeTypeServiceInterface::class);
+
+        // attribute.core.string is already used by StringAttribute, so when we try to register
+        // ExampleCustomAttributeWithUsedName (which has the same ID) we should get this exception
+        $this->expectException(AttributeTypeAlreadyRegisteredException::class);
```
stu-greenshoots, line 72: This falls into the pattern of other services we provide, but should we actually allow someone to overwrite an AttributeType?

elliotmassen: I think that could well be interesting, specifically once we build out the plugin system. I'm thinking for now it'd be better to not allow it and then open it up later, as opposed to opening it up now and trying to walk it back later.

src/AttributeEngine/Tests/AttributeResolverServiceTest.php
```diff
@@ -41,17 +41,42 @@ public function testAccessToUnsupportedAttribute()
...
+            'opendialog.attribute_engine.custom_attribute_types',
+            [ExampleCustomAttributeType::class]
+        );
+
+        $this->setConfigValue(
+            'opendialog.attribute_engine.custom_attributes',
+            ['test_attribute' => ExampleCustomAttributeType::class]
+        );
+
+        // Our custom attribute type isn't register so we should fallback to a string attribute
```
bendeeson, line 75: Should we fallback to a string attribute? The attributes types are written in code, and manually registered in config. It feels like it should be 'developer error' if they're note registered. Should it be a fatal exception instead?

bendeeson (commented):
Overall looks good to me. Just one comment with a question regarding how we handle unregistered attribute types.

src/AttributeEngine/Tests/AttributeResolverServiceTest.php
elliotmassen: Ah this is a good point! 👍  I was leaning on the current behaviour described in `AttributeResolverServiceTest.testBadBinding`, but I agree I think an exception is the way to go here. I'll commit something for this shortly.

bendeeson (approved):

PR #433, opened 2021-02-02
Dynamic attributes

bendeeson:
Introduces the 'Dynamic attributes' concept. Dynamic attributes are (attribute id, attribute type) pairs stored in the database. They allow for defining attributes outside of configuration files.
This PR introduces a new DynamicAttribute laravel model to the attribute engine and integrates it into the attribute resolver.


src/AttributeEngine/Facades/AttributeResolver.php
```diff
@@ -9,6 +9,10 @@
  * @method static AttributeInterface getAttributeFor(string $attributeId, mixed $value)
  * @method static AttributeInterface[] getSupportedAttributes()
  * @method static void registerAttributes($attributes)
+ * @method static void registerAllDynamicAttributes()
```
elliotmassen, line 12: Let's move this functionality into the `AttributeEngineServiceProvider`, the resolver should be fairly agnostic to the source of attributes. I'll leave further comments on the service provider file.

src/AttributeEngine/Tests/AttributeResolverServiceTest.php
```diff
@@ -2,13 +2,19 @@
...
 use OpenDialogAi\AttributeEngine\AttributeResolver\AttributeResolver;
+use OpenDialogAi\AttributeEngine\Attributes\IntAttribute;
 use OpenDialogAi\AttributeEngine\Attributes\StringAttribute;
 use OpenDialogAi\AttributeEngine\Exceptions\UnsupportedAttributeTypeException;
+use OpenDialogAi\AttributeEngine\DynamicAttribute;
 use OpenDialogAi\Core\Tests\TestCase;
 
 class AttributeResolverServiceTest extends TestCase
 {
+    use RefreshDatabase;
```
elliotmassen, line 16: I think our test suite should be doing this automatically before each test - did you encounter an issue relating to this?

elliotmassen (changes requested):
Thanks for this! Please see my comments below regarding rethinking the registration of the dynamic attributes to be more consistent with existing registrations.

src/AttributeEngine/AttributeEngineServiceProvider.php
```diff
@@ -31,6 +31,8 @@ public function register()
             if (is_array(config('opendialog.attribute_engine.custom_attributes'))) {
                 $attributeResolver->registerAttributes(config('opendialog.attribute_engine.custom_attributes'));
             }
+            // Gets custom dynamic attributes from the database
+            $attributeResolver->registerAllDynamicAttributes();
```
elliotmassen, line 35: I think we should keep this consistent with existing attribute registration (as seen with core and app attributes above), with something similar to:

```suggestion
            if ($this->hasDynamicAttributes()) {
                $attributeResolver->registerAttributes($this->formatDynamicAttributes());
            }
```

This way the resolver remains fairly agnostic to the source of registered attributes. The outcome of this change is to create some `formatDynamicAttributes` method that can retrieve dynamic attributes from the database and map them the same format as found in the config files:

```php
[
    'my_custom_str_attr' => StringAttribute::class,
    'my_custom_bool_attr' => BooleanAttribute::class,
    ...
]
```

For this we can make use of `AttributeTypeService.getAttributeTypeClass` to turn attribute type IDs (eg. `attribute.core.string`) into attribute type classes (eg. `StringAttribute::class`).

src/AttributeEngine/AttributeResolver/AttributeResolver.php
```diff
@@ -56,8 +83,29 @@ public function registerAttributes(array $attributes): void
...
+                continue;
+            }
+            if ($this->attributeTypeService->isAttributeTypeAvailable($dynamicAttribute->attribute_type)) {
+                $attributeTypeClass = $this->attributeTypeService->getAttributeTypeClass($dynamicAttribute->attribute_type);
+                $this->supportedAttributes[$dynamicAttribute->attribute_id] = $attributeTypeClass;
+            } else {
+                Log::error(sprintf("Not registering dynamic attribute %s - has unknown attribute type identifier %s",
+                    $dynamicAttribute->attribute_id, $dynamicAttribute->attribute_type));
+            }
+        }
```
elliotmassen, line 102: Some of these checks already exist in the `registerAttributes` method above, and others would be better placed in the service provider. Please see my comment in the service provider file for how we can look to remove this method.

elliotmassen (changes requested):
Ah my last review lost some of my comments - I've rewritten them and added to this review.

src/AttributeEngine/AttributeEngineServiceProvider.php
bendeeson: This has been updated.

elliotmassen (approved):
This ready to merge, but I'd imagine we'll want to get #435 merged first, as it contains parts of #434 as does this PR.

PR #434, opened 2021-02-03
Reflection API

elliotmassen:
This PR introduces a `ReflectionHelper` which provides reflection data about all OpenDialog components (and is expose to the administrator UI by https://github.com/opendialogai/opendialog/pull/238).

## OpenDialog Components

A major part of this work is the introduction of the `OpenDialogComponent` contract. This covers the common data between our various components: ID, type, source, human-readable name & human-readable description. Much of the changes cover rolling out this contract and its implementation across the components. A specific change is related to the subsumption of `HasName` trait & `$name` property into `$componentId`.

### Property definitions

- `$componentId`: A unique name for the component, eg. `eq` operation or `interpreter.core.luis` interpreter
- `$componentType`: The type of component (see `ODComponentTypes` for a valid list), eg. `sensor` or `attribute`
- `$componentSource`: The type of component (see `ODComponentSource` for a valid list), eg. `core` or `app`. See below for more details.
- `$componentName`: The human-readable name for the component.
- `$componentDescription`: The human-readable name for the component.

### Component sources

Currently three valid component sources are defined `core`, `app` & `custom`. These are best explained by thinking about attributes. Core attributes are those defined by OpenDialog core and are necessary for the functioning of the conversation and periphery engines. App attributes are those defined in an applications `opendialog/attribute_engine.php` configuration file and will likely be necessary for the functioning of application-level components and conversational conditions. Custom attributes are those defined in the administrator UI (sometimes referred to as "dynamic" attributes), these are more likely to be used to collect outputs from actions or interpreters and to add dynamic content to messages.

Other component types currently only make use of `core` and `app`. In the future it seems conceivable that there may be a `plugin` (or similar) source to differentiate components provided by external code, or that if we had some kind of lambda action functionality that we may have actions with `custom` source.

## Changes to attribute registration

Attributes are the one component type that does not neatly fit into the `OpenDialogComponent` pattern. Simply put, the reason for this is that an attribute does not have a specific concrete class of its own. For example, it we define a first name attribute like so, we only have a class for the attribute **type** but not *the* attribute itself:

```php
[
    'first_name' => StringAttribute::class,
    ...
]
```

To account for this, an `AttributeDeclaration` class has been introduced to capture the various data required for attributes as components. Currently this is only necessary to capture the source of the attribute, but this declaration object may offer further uses in the future. It's worth noting that `AttributeDeclaration`'s are only used within the `AttributeResolver` and `AttributeEngineReflection` and therefore their introduction does not affect any other engines.

## Changes to operations

As it was necessary to provide information on the required attributes and parameters for operations there have been changes to how operations are defined. Operations no longer need to define a `getAllowedParameters` as it has been replaced with a static `requiredParametersArgumentNames` array property. If no property is provided it will default to presuming a required `value` parameter argument.

Similarly, there is now a static `requiredAttributeArgumentNames` array property which can be used to define the required attributes for an operation. If no property is provided it will default to presuming a required `attribute` attribute argument.

This means that if an application has a custom operation that requires an input attribute argument called `attribute` and an input parameter argument called `value`, then you will only need to remove the `getAllowedParameters` (unless you wish to explicitly restate the defaults). This is demonstrated throughout our core operations which mostly use these two input arguments throughout.

## Changes to contexts

In line with the introduction of components, the way contexts are defined has become stricter. Previously it was possible to programmatically create a new context with `ContextSerivce::createContext('my_custom_context')`. As components require static properties, this programmatic creation is no longer supported and therefore `BaseContext` has also been removed. If you require a custom context you can create one by extending `AbstractCustomContext` and registering your context in the context engine configuration.

Contexts can also define an `$attributesAreReadOnly` boolean property which signifies if it should be possible to alter attributes in the context. For `_intent` and `message_history` contexts we need OpenDialog to manage them, and therefore these are the two core contexts with `$attributesAreReadOnly` set to `true`. Currently this property is here for the benefit of the UI only, but in the future we could use the concept to enforce rules on context usage (perhaps by introducing a `ReadOnlyContext` class or similar).


istos:
@elliotmassen I went ahead and merged this in my attribute-1x branch so that I can work on top of what you have. Apologies for the heavy-handedness of it but This should avoid a world of pain for you trying to merge later. Once you've checked the changes in attribute-1x you can then branch of that to continue work and we could close this.

src/AttributeEngine/AttributeResolver/AttributeResolver.php
```diff
@@ -22,20 +23,20 @@ class AttributeResolver
     public static $validIdPattern = "/^([a-z]+_)*[a-z]+$/";
     public static $validTypePattern = "/^attribute\.[A-Za-z]+\.[A-Za-z_]+$/";
 
-    /* @var array */
-    private $supportedAttributes = [];
+    /** @var Map|AttributeDeclaration */
```
istos, line 26: Now that we have introduced Laravel Collections in the conversation model I would default to using collections anywhere we might think of using Map - especially in places where we were not using Map already.

src/AttributeEngine/AttributeResolver/AttributeDeclaration.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+
+namespace OpenDialogAi\AttributeEngine\AttributeResolver;
+
+
+use OpenDialogAi\AttributeEngine\Contracts\Attribute;
+
+class AttributeDeclaration
```
istos, line 9: Some comment here explaining what an AttributeDeclaration is.

src/ContextEngine/Contexts/User/UserContext.php
```diff
@@ -2,26 +2,27 @@
...
+
     public const USER_CONTEXT = 'user';
     public const UTTERANCE_USER = 'utterance_user';
 
     protected $persistentAttributes = [
         UserAttribute::CURRENT_USER
     ];
 
-    public function __construct(ContextDataClient $dataClient)
+    public function __construct(UserDataClient $dataClient)
```
istos, line 24: I am not sure that the UserContext needs to be aware of the fact that it is using anything other that a ContextDataClient - this will save us from introducing customized functionality in the UserDataClient that then the UserContext becomes dependent on.

istos (changes requested):
All looks good - my only blocking question is around making the UserContext dependent on the UserDataClient rather than a more generic ContextDataClient contract.

src/ContextEngine/Contexts/User/UserContext.php
elliotmassen: Ah yes, my bad I should've explained why I did this. While I was doing this work I figured that the Context engine could be standardised a bit and made it so that even core contexts are registered via a configuration array:

https://github.com/opendialogai/core/blob/ad31fa96c2dd654f9e371b2411fd5a0b85c530d6/src/ContextEngine/ContextEngineServiceProvider.php#L23-L33

This removed the ability to directly pass parameters to the constructor, and therefore the solution was to lean on Laravel's dependency injection here. As we're using `resolve()` it will automatically pass in the `UserDataClient` for us. But to be able to do this it requires the bound class name to resolve against, hence the change to `UserDataClient`. I was thinking that as the parent `PersistentContext` constructor still retains the more general `ContextDataClient` that it still gives us that flexibility.

https://github.com/opendialogai/core/blob/ad31fa96c2dd654f9e371b2411fd5a0b85c530d6/src/ContextEngine/ContextService/CoreContextService.php#L111-L115

I definitely take your point though, and I'm happy to revert the change if you'd like. My argument in favour of keeping it (perhaps with adjustments) is that we will likely need to be able to parameterise the constructors of custom contexts which will require the same considerations.

istos: Got it. Ok - let's leave it as is for now. Might make sense to get a feel for what else might come up and whether we do need to do something different in the future 👍

istos (approved):

PR #435, opened 2021-02-05
OpenDialog 1.x Refactor

istos:
Attributes in previous versions of OpenDialog were used inconsistently, this work aims to make attribute definitions more robust in preparation for applying attributes as the main unit of data exchange between all the various components. 

### Attribute Engine Contracts

AttributeEngine gets a Contracts namespace with all the core contracts for the engine

- Attribute: any describable feature of the environment (e.g. price, color, name, postcode, etc)
- AttributeValue: the specific type of value that is used to describe an attribute (e.g. string, int, float, array, etc)
- ScalarAttribute: an attribute that only requires one value to be described
- CompositeAttribute: an attribute that is described through other attributes (user.name, user.dob, etc)
- CollectionAttribute: an attribute that is described through a collection of AttributeValues  _contract in place but implementation lacking_

### AttributeResolver

The AttributeResolver should be the main point through which Attribute objects are created. The function signature has changed to accommodate a couple of different ways of using it. 

`    public function getAttributeFor(string $attributeId, $rawValue = null, AttributeValue $value = null)
`
The most often used way will still be to send an `attributeId` (which should be a registered id through config) and a `rawValue` - the Attribute object implementation will then deal with the `rawValue` by wrapping it (where possible) in an AttributeValue object.

### Serializing and deserializing attributes

Attributes *and* AttributeValues should implement their own `jsonSerialize()` to ensure consistent serialization and deserialization of attributes, especially for persistence. 

#### IncomingWebchat Controller 
The incoming webchat controller was moved out. of the SensorEngine to enable the SensorEngine to only depend. on AttributeEngine and make testing separate components simpler

#### UtteranceAttribute
Sensors should generate `UtteranceAttribute`s, which for the time being I've kept as a single UtteranceAttribute in AttributeEngine with some helper functions. In the future, we can decide to move that functionality to a more specific `WebchatUtteranceAttribute`
The _old_ `Utterance` namespace is still there but will be removed. 

## Tests

Only tests passing are withing the AttributeEngine and SensorEngine packages, the rest will come online as I work through the refactor.


src/ActionEngine/Actions/BaseAction.php
```diff
@@ -3,22 +3,29 @@
...
 
 use Ds\Map;
+use OpenDialogAi\Core\Components\BaseOpenDialogComponent;
+use OpenDialogAi\Core\Components\Contracts\OpenDialogComponent;
+use OpenDialogAi\Core\Components\ODComponent;
+use OpenDialogAi\Core\Components\ODComponentTypes;
 use OpenDialogAi\Core\Traits\HasName;
 
-abstract class BaseAction implements ActionInterface
+abstract class BaseAction implements ActionInterface, OpenDialogComponent
```
elliotmassen, line 12: Is there any reason why `ActionInterface` couldn't extend `OpenDialogComponent`? In my previous work I had it mapped like: concrete class extends abstract class, abstract class implements interface, interface extends OD component interface.

src/AttributeEngine/AttributeValues/BooleanAttributeValue.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+
+namespace OpenDialogAi\AttributeEngine\AttributeValues;
+
+class BooleanAttributeValue extends AbstractAttributeValue
+{
+    public static $attributeValueType = 'attributeValue.core.boolean';
```
elliotmassen, line 8: Is there any chance we could use `attribute_value` rather than `attributeValue`? This would keep the casing consistent - ie. we use `attribute_type` for attribute types.

src/AttributeEngine/AttributeResolver/AttributeResolver.php
```diff
@@ -74,15 +78,34 @@ public function isAttributeSupported(string $attributeId)
      *
      * @param string $attributeId
      * @param $value
-     * @return AttributeInterface
+     * @return Attribute
      */
-    public function getAttributeFor(string $attributeId, $value)
+    public function getAttributeFor(string $attributeId, $rawValue = null, AttributeValue $value = null): Attribute
```
elliotmassen, line 83: I'm not sure I understand why the signature here could not be:

```suggestion
    public function getAttributeFor(string $attributeId, $value = null): Attribute
```

Then within the method we could condition on `$value instanceof AttributeValue` to determine how to handle it. It feels like keeping the signature intuitive would be preferable, as this is a method that non-core developers may use in custom contexts or actions for example.

Perhaps I'm missing a piece of the puzzle here?

src/ContextEngine/Contexts/User/UserContext.php
```diff
@@ -2,266 +2,52 @@
...
+     * @inheritdoc
      */
-    public function getAttributeValue(string $attributeName)
+    public function getAttribute(string $attributeName): Attribute
     {
-        if ($this->userService->hasUserAttribute($this->getUser(), $attributeName)) {
-            $userAttribute = $this->userService->getUserAttributes($this->getUser())->get($attributeName);
-            return $userAttribute->getInternalAttribute()->getValue();
+        if ($attributeName == self::CURRENT_USER) {
+            return $this->getUser();
```
elliotmassen, line 31: Would it be possible to move this to the constructor (or similar `loadAttributes ` method)? It would be great to allow classes that use `HasAttributesTrait` to manage attributes properly - in this case `current_user` becomes harder to work with as it's got it's own custom getting behaviour. If it would be possible to do something like the following in the constructor (or similar) I think it might make the use of the context cleaner:

```php
$currentUser = $this->getUser();
$this->addAttribute(self::CURRENT_USER, $currentUser);
```

If the above is done we won't have a need to override `getAttribute`.

elliotmassen (commented):
I've just spent 15 minutes skimming through and pulling up my immediate more conceptual concerns/comments:

src/AttributeEngine/AttributeValues/BooleanAttributeValue.php
istos: Done.

src/AttributeEngine/AttributeResolver/AttributeResolver.php
istos: Absolutely - done.

src/ContextEngine/Contexts/User/UserContext.php
istos: My concern with moving it to the constructor is that we will end up loading everything upfront - which is the problem with the `loadAttributes` method as well. The moment you create the context you will be loading a user but you might not know what user you should be creating then. We are still calling the parent function so functionality remains consistent other than where it differs. 

More broadly the reason contexts exist in general is so that we can delegate work to them around how to handle the harder aspects of attributes (especially where attributes carry specific semantic meaning such as `current_user`. To me that overwrite exposes the fact that some attributes are different and should be dealt with different while sticking a balance with consistent behavior elsewhere.

stu-greenshoots: Is there a path with deferred loading - so contexts are not loaded into the service layer until they are directly used, and at that point, the `loadAttributes` method is called loading of the attributes in a standard manner. This would only give the benefit if all of the context data can be loaded from a single place I suppose.

I think this would also mean having a generic 'User' context that could be used for any user data, and an instantiation of that user context for the current user

We have experienced pain with how contexts currently work with differences between what can be saved and what can be retrieved, so I feel where Elliot is coming from

Tests/TestCase.php
```diff
@@ -78,18 +73,18 @@ public function getPackageProviders($app)
     {
         return [
             CoreServiceProvider::class,
-            GraphServiceProvider::class,
             ActionEngineServiceProvider::class,
-            ConversationBuilderServiceProvider::class,
+//            ConversationBuilderServiceProvider::class,
```
stu-greenshoots, line 77: Could these be taken as an input to the setup method so that each module can be very specific about what it needs to boot?

We could even have a base test case per module that extends this one to enforce the separation

Tests/Unit/ODControllerTest.php
```diff
@@ -12,5 +15,21 @@ public function testBinding()
...
+
+        $utterance = $this->createWebchatMessageUtteranceAttribute();
+        $od->runConversation($utterance);
+
+    }
+
+    public function testConnectToDgraph()
+    {
+        $dataClient = $this->app->make(ConversationDataClient::class);
+        dump($dataClient->query());
```
stu-greenshoots, line 36: does this need to validate anything?

src/ActionEngine/Actions/BaseAction.php
```diff
@@ -3,22 +3,29 @@
...
+    protected static string $componentSource = ODComponentTypes::APP_COMPONENT_SOURCE;
 
     protected static $name = 'action.core.base';
 
-    /** @var string[] */
-    protected $requiredAttributes = [];
+    /** @var array|string[] */
+    protected static $requiredAttributes = [];
 
     /** @var Map */
```
stu-greenshoots, line 24: Why is the input a map, but the output an array?

src/AttributeEngine/AttributeTypeService/AttributeTypeService.php
```diff
@@ -96,7 +98,22 @@ public function registerAttributeTypes(array $attributeTypes): void
...
+                // OpenDialog Component class by looking for the interface implementation. The string could
+                // not refer to valid class which will throw the first generic exception/
+                try {
+                    $interfaces = class_implements($attributeType);
+                } catch (\ErrorException $e) {
+                    throw new UnsupportedAttributeTypeException();
+                }
+
+                // Having ensured it is a class we now check that it is an OpenDialogCompoment class.
+                if (in_array('OpenDialogAi\Core\Components\Contracts\OpenDialogComponent', $interfaces)) {
```
stu-greenshoots, line 111: Could this user OpenDialogComponent::class rather than a string?

src/AttributeEngine/AttributeTypeService/AttributeTypeService.php
```diff
@@ -96,7 +98,22 @@ public function registerAttributeTypes(array $attributeTypes): void
     {
         foreach ($attributeTypes as $attributeType) {
             try {
-                $this->registerAttributeType($attributeType);
+                // Check that the attributeType is a valid class and subsequently ensure it is a valid
+                // OpenDialog Component class by looking for the interface implementation. The string could
+                // not refer to valid class which will throw the first generic exception/
+                try {
+                    $interfaces = class_implements($attributeType);
+                } catch (\ErrorException $e) {
```
stu-greenshoots, line 106: Could an exception be thrown here? I think it could return false instead

stu-greenshoots: Maybe something like:
```if (!$interfaces = class_implements($attributeType)) {
    throw new UnsupportedAttributeTypeException();
}```

File: src/AttributeEngine/AttributeResolver/AttributeResolver.php (hunk 3 above)
Comment (elliotmassen) [reply src/AttributeEngine/AttributeResolver/AttributeResolver.php:83]: Great :) I see you made changes relating to attribute constructors regarding this, did you also intend to do the same for `getAttributeFor`?

File: Tests/Unit/ODControllerTest.php (hunk 6 above)
Comment (istos) [reply Tests/Unit/ODControllerTest.php:36]: Definitely - this is just a placeholder for future work.

File: Tests/TestCase.php (hunk 5 above)
Comment (istos) [reply Tests/TestCase.php:77]: Yes, I think a review of our TestCase base setup is worth doing. I would move that to a separate subsequent step though.

File: src/ContextEngine/Contexts/User/UserContext.php (hunk 4 above)
Comment (istos) [reply src/ContextEngine/Contexts/User/UserContext.php:31]: Feels like there is space for improvement here but I would defer to a later step again.

File: src/Conversation/ConversationObject.php [hunk 10]
```diff
@@ -0,0 +1,206 @@
+<?php
+namespace OpenDialogAi\Core\Conversation;
+
+class ConversationObject
+{
+    public const UNDEFINED = 'undefined';
+
+    protected string $type = self::UNDEFINED;
+    protected string $odId = self::UNDEFINED;
```
elliotmassen, line 9: Is there any reason why this is called `odId` as opposed to just `id`? I'm wondering if that might be preferable for consistency (eg. we use `name` rather than `odName`, `description` rather than `odDescription`).

src/Conversation/ODObjectCollection.php
```diff
@@ -0,0 +1,47 @@
...
+    public function addObject(ConversationObject $object)
+    {
+        $odId = $object->getODId();
+        if (!isset($odId) || $odId == ConversationObject::UNDEFINED) {
+            throw new CannotAddObjectWithoutODidException();
+        }
+        $this->put($object->getOdId(), $object);
+    }
+
+    public function getObject($odId)
```
elliotmassen, line 23: Does this account for multiple objects with the same ID? ie even within a conversation/scene/turn we could expect to have multiple intents with the same ID. This is the issue that gave rise to `hashOrId` in core v0:

https://github.com/opendialogai/core/blob/3290352ce31ee862b885f424a9880d429678d3ac/src/Graph/Node/Node.php#L303-L310

src/InterpreterEngine/Interpreters/NoMatchIntent.php
```diff
@@ -13,7 +13,5 @@ class NoMatchIntent extends Intent
 
     public function __construct()
     {
-        parent::__construct(self::NO_MATCH);
-        parent::setConfidence(1);
     }
```
elliotmassen, line 16: I'm wondering if this could stay? It's certainly useful in custom interpreters.

elliotmassen (commented):
Here's my brief thoughts on issues that seem more major or that might cause us a bit of pain moving forward.

src/InterpreterEngine/Interpreters/NoMatchIntent.php
elliotmassen: Resolving this as it seems the upcoming conversation engine refactor will re-add something similar to this.

src/Conversation/ConversationObject.php
istos: The thinking was that this reminds everyone that this is not a "normal" id that you would find on any model (usually unique, etc) - it is the opendialog CDL id.

PR #440, opened 2021-02-15
Conversation Engine

istos:


src/ContextEngine/Contexts/PersistentContext.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        if (!$this->isPersistentAttribute($attributeName)) {
+            return parent::getAttribute($attributeName);
+        }
+
+        if ($this->hasAttribute($attributeName) && !$refresh) {
+            return parent::getAttribute($attributeName);
+        } elseif (($this->hasAttribute($attributeName) && $refresh) || (!$this->hasAttribute($attributeName))) {
+            try {
+                $this->dataClient->loadAttribute($attributeName);
```
elliotmassen, line 45: Is there a missing bit here where, after loading the attribute, it is added to the internal attribute map and returned? Something like:
```suggestion
                $attribute = $this->dataClient->loadAttribute($attributeName);
                this->attributes->put($attribute->getId(), $attribute);
                return $attribute;
```

src/ConversationEngine/ConversationEngine.php
```diff
@@ -51,17 +58,112 @@ public function setActionEngine(ActionEngineInterface $actionEngine): void
...
+        );
+        ContextService::saveAttribute(
+            self::CONVERSATION_CONTEXT.Intent::CURRENT_INTENT,
+            $intent->getODId()
+        );
+        ContextService::saveAttribute(
+            self::CONVERSATION_CONTEXT.Intent::CURRENT_SPEAKER,
+            $intent->getSpeaker()
+        );
+    }
```
elliotmassen, line 137: This is a really neat use of contexts, and breathes a bit of purpose into the conversation context 👍 

I think it might be a little tidier (and perhaps a better precedent to set?) to do something more like the following, which is the way I'd usually advise use of the `ContextService`.

```php
$conversationContext = ContextService::getContext(self::CONVERSATION_CONTEXT);
$conversationContext->addAttribute(Scenario::CURRENT_SCENARIO, $intent->getScenario()->getODId());
$conversationContext->addAttribute(Conversation::CURRENT_CONVERSATION, $intent->getConversation()->getODId());
...
```

src/ConversationEngine/Reasoners/ConversationalStateReasoner.php
```diff
@@ -4,22 +4,80 @@
...
+                // the history record (which is just a snapshot in time) we can use the conversation context
+                // to track where our "current thinking" is. Once a reasoning cycle is done we then extract
+                // information from the conversation context to put it back into a history record.
+                ContextService::saveAttribute(
+                    self::CONVERSATION_CONTEXT.Scenario::CURRENT_SCENARIO,
+                    $record->getScenarioId()
+                );
+                ContextService::saveAttribute(
+                    self::CONVERSATION_CONTEXT.Conversation::CURRENT_CONVERSATION,
+                    ConversationObject::UNDEFINED
```
elliotmassen, line 46: I'd echo my previous comment on use of `ContextService` throughout here.

src/ConversationEngine/Reasoners/IntentFilter.php
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        $matchingIntents = $interpretedIntents->filter(function (Intent $interpretedIntent, Intent $intent) {
+            if ($interpretedIntent->getODId() == $intent->getODId()) {
+                // Set the confidence value
+                $intent->setConfidence($interpretedIntent->getConfidence());
+                // Get any attributes that were identified by the interpreter
+                $intent->setAttributes($interpretedIntent->getAttributes());
+            }
+        });
```
elliotmassen, line 40: This looks to be missing a `return` statement, and some finishing in general.

elliotmassen (commented):
Just a few comments. I've made note of a less-pressing optimisation for condition filtering in the refactor notes for future reference.

src/ContextEngine/Contexts/User/UserContext.php
```diff
@@ -3,51 +3,25 @@
...
 {
     public const USER_CONTEXT = 'user';
-    public const CURRENT_USER = 'current_user';
     public const UTTERANCE_USER = 'utterance_user';
-    public const USER_ID_SCOPE = 'user_id';
 
-    protected UserDataClient $dataClient;
+    protected $persistentAttributes = [
+        UserAttribute::CURRENT_USER
+    ];
```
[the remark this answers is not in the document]
istos: Agree. Will leave as is for now but definitely space to do something else  - I was missing the understanding of how attributes are loaded in custom contexts.

src/ContextEngine/Contexts/PersistentContext.php
istos: Good catch :-) added a fix for this.

src/ConversationEngine/ConversationEngine.php
istos: It is definitely tidier - the main concern I have is that we are adding attributes without going through the attribute resolver like this. `ContextServer::saveAttribute` provides a centralised, standardised point that also uses the attribute resolver. `$context->addAttribute` bypasses that check so would allow you to add attributes that are not actually properly configured.

PR #443, opened 2021-02-16
Starting Intent Selector

istos:
This adds a more generic IntentInterpreter filter in our reasoners and uses a new technique for first interpreting intents and then looking for matching intents. 

We now store interpretations of intenta from an interpreter withing the original intent. 

```
$intent->addInterpretedIntents($interpretations);
```

This will allow us to do more fine-grained comparisons between what we were looking for and what actually happened. 

Intents with interpreter intents can then check for matches themselves. 

```
    public function checkForMatch():bool
    {
        /* @var Intent $intent */
        foreach ($this->interpretedIntents as $intent) {
            if (($intent->getODId() == $this->getODId()) && ($intent->getConfidence() >= $this->getConfidence())) {
                // @todo currently this will give us the last matching intent out of a number of intents as the "winner"
                // so we may want to consider ranking within a single intent as well if we got multiple matches.
                $this->interpretation = $intent;
                return true;
            }
        }
        return false;
    }
```

This will also allow us to deal with attributes coming out of interpreters (that will be stored against the `interpretedIntent` collection within an Intent and the attributes that the Intent itself was expecting as per the CDL (nb: that work is not here)


src/Conversation/Intent.php
```diff
@@ -67,6 +73,43 @@ public function setConfidence(float $confidence)
...
+        foreach ($this->interpretedIntents as $intent) {
+            if (($intent->getODId() == $this->getODId()) && ($intent->getConfidence() >= $this->getConfidence())) {
+                // @todo currently this will give us the last matching intent out of a number of intents as the "winner"
+                // so we may want to consider ranking within a single intent as well if we got multiple matches.
+                $this->interpretation = $intent;
+                return true;
+            }
+        }
+        return false;
+    }
```
elliotmassen, line 102: Just wanted to make note of what we discussed earlier here. It seems for now this method does what it needs to (in that it returns an boolean represent it one of the interpreted intents matches. The tricky bit is the use of `$this->interpretation` as this presumes at this point a _single_ match can be chosen. This skips over evaluating an conditions, and is perhaps better chosen later.

src/ConversationEngine/ConversationEngine.php
```diff
@@ -92,7 +93,7 @@ public function getNextIntents(UtteranceAttribute $utterance): IntentCollection
             if ($responseIntents->isEmpty()) {
                 return $this->noMatchCollection();
             }
-        } catch (AttributeDoesNotExistException $e) {
+        } catch (CouldNotCreateUserFromUtteranceException $e) {
```
elliotmassen, line 96: Could this try/catch could just be wrapped around `$currentUser = $this->getCurrentUser($utterance);` rather than the whole method body? Ie. are we expecting any other lines to throw `CouldNotCreateUserFromUtteranceException` besides the call to `getCurrentUser`?

src/ConversationEngine/Reasoners/StartingIntentSelector.php
```diff
@@ -14,16 +19,17 @@ class StartingIntentSelector
...
-            ConditionFilter::checkConditions($intent);
+            return ConditionFilter::checkConditions($intent);
         });
 
-        $interpreterMatchingIntents = $intents->filter(function ($intent) {
-            IntentFilter::matchIntent($intent);
-        });
+        // Now we can pass each intent through interpreters and interpret given the utterance
+        $utterance = ContextService::getAttribute(UtteranceAttribute::UTTERANCE, UserContext::USER_CONTEXT);
+        return IntentInterpreterFilter::filter($conditionPassingIntents, $utterance);
```
elliotmassen, line 32: The ordering of condition evaluation and interpretation is the reverse of what it'll need to be for incoming intents. For incoming intents we must interpret first, then evaluate conditions (as we won't have access to the `_intent` context attributes, i.e the interpreted entities, until interpretation). You may remember this became apparent when we adding conditions to incoming intents (primarily described in https://github.com/opendialogai/core/pull/256). If you consider the intention of setting conditions on an incoming intent, it is to filter the possible matching intents after interpretation, not prior to it. It is to say that if interpretation filtered the number of possible matching intents from 5 down to 3, then condition evaluation is likely to filter the number from 3 to 1.

src/AttributeEngine/AttributeResolver/AttributeResolver.php
```diff
@@ -92,14 +92,22 @@ public function getAttributeFor(string $attributeId, $value = null): Attribute
             if ($attribute instanceof ScalarAttribute) {
                 if ($value instanceof AttributeValue) {
                     $attribute->setAttributeValue($value);
+                } elseif ($value instanceof ScalarAttribute) {
+                    $attribute->setRawValue($value->getAttributeValue()->getRawValue());
```
elliotmassen, line 96: What's the significance of this change? I'm struggling to understand why `$value` should ever be an attribute.

elliotmassen (commented):
This looks good - and as we said earlier gives a general shape to the work that is outstanding. I've added some comments about incoming condition evaluation which I think would necessitate a bit of a refactor. I'll leave it with you to say whether that's something to iron-out now, or later.

src/Conversation/Intent.php
istos: Indeed - I am not using that right now and I am sure we will revisit once we move outside of having to choose just opening intents.

src/ConversationEngine/Reasoners/StartingIntentSelector.php
istos: Good point I've changed the order and added a ticket to capture the task of populating the intent context with outputs from Interpreters and having some tests around that.

src/AttributeEngine/AttributeResolver/AttributeResolver.php
istos: The attribute resolve is attempting to always do the best possible thing for whoever is trying to resolve an attribute. It is possible that a context is trying to add an attribute to itself.

src/ConversationEngine/ConversationEngine.php
istos: Yes we should eventually do that - for now it is slightly artificial to get the most basic run through the conversation engine going.

PR #445, opened 2021-02-19
ActionPerformer

elliotmassen:
This PR adds an `ActionPerformer` class that:

 - defines methods that take input of `Intent`(s) or `Action`(s),
 - provide the actions with the desired attribute inputs,
 - perform the associated actions,
 - take the actions result attributes and store them in the desired contexts,
 - and persists any updated persistent contexts.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -66,9 +66,10 @@
             <directory suffix="Test.php">./src/SensorEngine/Tests</directory>
         </testsuite>
 
-<!--    <testsuite name="ConversationEngine">
+        <testsuite name="ConversationEngine">
             <directory suffix="Test.php">./src/ConversationEngine/Tests</directory>
-        </testsuite> -->
+            <exclude>./src/ConversationEngine/Tests/ConversationEngineTest.php</exclude>
+        </testsuite>
 
         <testsuite name="NlpEngine">
             <directory suffix="Test.php">./src/NlpEngine/Tests</directory>
--- a/src/ConversationEngine/Reasoners/ActionPerformer.php
+++ b/src/ConversationEngine/Reasoners/ActionPerformer.php
@@ -17,9 +17,21 @@ use OpenDialogAi\ContextEngine\Facades\ContextService;
 use OpenDialogAi\Core\Conversation\Action;
 use OpenDialogAi\Core\Conversation\ActionsCollection;
 use OpenDialogAi\Core\Conversation\Intent;
+use OpenDialogAi\Core\Conversation\IntentCollection;
 
 class ActionPerformer
 {
+    /**
+     * Performs all actions associated with the given intents
+     *
+     * @param IntentCollection $intents
+     */
+    public static function performActionsForIntents(IntentCollection $intents): void
+    {
+        foreach ($intents as $intent) {
+            self::performActionsForIntent($intent);
+        }
+    }
     /**
      * Performs all actions associated with the given intent
      *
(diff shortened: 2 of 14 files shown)


src/ConversationEngine/Tests/ActionPerformerTest.php
```diff
@@ -0,0 +1,178 @@
...
+            'composite.first_name' => 'session'
+        ]));
+
+        ActionPerformer::performAction($action);
+
+        $this->assertEquals(
+            'MY_NAME',
+            ContextService::getAttributeValue('composite.first_name', SessionContext::getComponentId())
+        );
+    }
```
elliotmassen, line 55: This test is failing currently, I think it may require some related work in the context or attribute engine.

phpunit.xml
```diff
@@ -66,9 +66,10 @@
             <directory suffix="Test.php">./src/SensorEngine/Tests</directory>
         </testsuite>
 
-<!--    <testsuite name="ConversationEngine">
+        <testsuite name="ConversationEngine">
             <directory suffix="Test.php">./src/ConversationEngine/Tests</directory>
-        </testsuite> -->
+            <exclude>./src/ConversationEngine/Tests/ConversationEngineTest.php</exclude>
+        </testsuite>
```
elliotmassen, line 72: I've updated this more of the test suite configuration, to give us a way of iteratively expanding the ConversationEngine tests, while excluding more holistic tests until later.

src/ConversationEngine/Reasoners/ActionPerformer.php
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Performs the given action
+     *
+     * @param Action $action
+     */
+    public static function performAction(Action $action): void
+    {
+        $inputAttributes = $action->getInputAttributes();
+
+        $actionResult = resolve(ActionEngineInterface::class)->performAction(
```
istos, line 64: Could we do a Facade from the ActionEngineInterface so that we have a consistent way of accessing engines (always through Facades)

src/ConversationEngine/Reasoners/ActionPerformer.php
```diff
@@ -0,0 +1,137 @@
...
+            } catch (ContextDoesNotExistException $e) {
+                $context = ContextService::getSessionContext();
+                Log::debug(sprintf(
+                    "'%s' saved to session context as provided content '%s' is not registered.",
+                    $attributeId,
+                    $contextId
+                ));
+            }
+
+            $context->addAttribute($attribute);
```
istos, line 116: This is saving directly to the context and not going through the AttributeResolver which means we would be able to save attributes that are defined in the CDL of the app, but are not part of the supported attributes for the app?

istos (changes requested):
Looking good @elliotmassen 👍 - main thing for me would be the Facade to access that ActionEngineInterface implementation instead of the Laravel resolve() function

src/ConversationEngine/Reasoners/ActionPerformer.php
elliotmassen: Hmm this is a good point - currently as the `ActionInterface` enforces each action to implement a `perform` method that returns an `ActionResult` (which holds an array of `Attribute`s - see below for the common factory method we use in actions), we give this control of resolving the attribute to each action.

As you say this means that the action could be instantiating attributes directly as opposed to running them through the resolver. I'm not sure how we'd get round this until we have actions returning `AttributeValue`'s as opposed to `Attribute`'s.

https://github.com/opendialogai/core/blob/f7771ec62f3dbe2a2eca5059883ff5711f690ed6/src/ActionEngine/Actions/ActionResult.php#L37-L46

istos (approved):

PR #446, opened 2021-02-23
IntentRanker

elliotmassen:
This PR implements the `IntentRanker` class, which provides a `getTopRankingIntent` method. The `IntentRanker` loops through all given intents and picks the first intent that primarily has the highest confidence and secondarily the most attributes.


src/ConversationEngine/Reasoners/IntentRanker.php
```diff
@@ -0,0 +1,60 @@
...
+     * @param Intent $b
+     * @return bool
+     */
+    private static function isIntentHigherRanking(Intent $a, Intent $b): bool
+    {
+        $confidenceComparison = $a->getInterpretation()->getConfidence() <=> $b->getInterpretation()->getConfidence();
+
+        if ($confidenceComparison === 0) {
+            // If the confidences were the same judge by number of (matching expected) attributes
+            return count($a->getInterpretation()->getAttributes()) > count($b->getInterpretation()->getAttributes());
```
elliotmassen, line 54: @istos I'm making a presumption here that the attributes added to the "interpretation" intent are _only_ those the intent defined as expected.

istos: I think we might want to do something broader moving forward but for now let's go with that 👍

src/ConversationEngine/Reasoners/IntentRanker.php
```diff
@@ -0,0 +1,60 @@
...
+
+        if (count($intents) < 2) {
+            return $intents->first();
+        }
+
+        /** @var Intent $topRankingIntent */
+        $topRankingIntent = null;
+
+        /** @var Intent $intent */
+        foreach ($intents as $intent) {
```
istos, line 32: Have you considered using the sortBy() method so that we first have a sorted view of intents and from that we can then extract the top ranking one. That also allows us to be more flexible in our sorting strategies in the future.

elliotmassen: I definitely agree that would be ideal. I did try using it but I wasn't able to get it to order the intents in a way that respected all our conditions (1. confidence, 2. attributes, 3. initial ordering). That's why I opted for the more manual approach of performing the selection of the top ranking intent during iteration, as it allow me to take the first one by default, then evaluate each subsequent intent relative to our current top ranking intent and replace it is of higher rank. The plus is that the tests most definitely capture the three necessary ranking conditions so we are able to amend this method in future with confidence.

istos (approved):

PR #448, opened 2021-02-26
Feature/graphql client

bendeeson:
Introduced a new 'GraphQL client' package.
This package includes a small DGraph GraphQL client that wraps Guzzle with some additional error checking as well as commands for setting the graphql schema and for wiping the the dgraph database.


config/opendialog.php
```diff
@@ -2,9 +2,7 @@
 
 return [
     'PACKAGE_NAME' => env('PACKAGE_NAME'),
```
bendeeson, line 4: Moved to GraphQLClient config

src/GraphQLClient/config/opendialog-graphql.php
```diff
@@ -0,0 +1,154 @@
...
+
+
+return [
+    'DGRAPH_URL' => env('DGRAPH_URL', function () {
+        throw new Exception("Missing env var DGRAPH_URL");
+    }), 'DGRAPH_PORT' => env('DGRAPH_PORT', function () {
+        throw new Exception("Missing env var DGRAPH_PORT");
+    }), 'DGRAPH_AUTH_TOKEN' => env('DGRAPH_AUTH_TOKEN', function () {
+        throw new Exception("Missing env var DGRAPH_AUTH_TOKEN");
+    }), "schema" => <<<'GQL'
```
bendeeson, line 11: This is the GraphQL schema to be passed to DGraph. It only contains types. DGraph generates the rest of the schema including data types etc from it.

.circleci/config.yml
```diff
@@ -33,18 +34,28 @@ jobs:
...
-          name: "Setup custom environment variables"
-          command: |
-            echo 'export OVERWRITE_DGRAPH_URL="http://localhost"' >> $BASH_ENV; echo 'export OVERWRITE_DGRAPH_PORT="8080"' >> $BASH_ENV
+          name: dgraph tunnel
+          background: true
+          command: socat TCP-LISTEN:8080,reuseaddr,fork "EXEC:'docker --tlsverify -H=$DOCKER_HOST --tlscacert=$DOCKER_CERT_PATH/ca.pem --tlscert=$DOCKER_CERT_PATH/cert.pem --tlskey=$DOCKER_CERT_PATH/key.pem exec -i dgraph_server_1 socat STDIO TCP-CONNECT:localhost:8080'"
 
       # run tests!
-      - run: phpdbg -qrr ./vendor/bin/phpunit --exclude skip --no-coverage -d memory_limit=8192M
+      - run: phpdbg -qrr ./vendor/bin/phpunit --exclude skip --no-coverage --configuration phpunit-ci.xml -d memory_limit=8192M
```
bendeeson, line 58: Uses a separate phpunit-ci.xml file now. This lets use remove the OVERWRITE_* env vars.

composer.json
```diff
@@ -25,11 +25,13 @@
     "barryvdh/laravel-ide-helper": "^2.7",
     "matthewbdaly/artisan-standalone": "^0.0.8",
     "dms/phpunit-arraysubset-asserts": "^0.2.1",
-    "friendsofphp/php-cs-fixer": "^2.15"
+    "friendsofphp/php-cs-fixer": "^2.15",
+    "opendialogai/dgraph-docker": "^20.11.0"
```
elliotmassen, line 29: I'm not savvy with the requirements for this work, so please treat this as a question (to help my own understanding) rather than feedback: Is there a reason why this is re-introduced? I had presumed that when it was [removed](https://github.com/opendialogai/core/pull/435/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34) in #435 that we would be mocking the client responses in tests and therefore we wouldn't require it as a dependency anymore. My understanding is definitely not 100% here, so I could well be missing something!!

bendeeson: I agree that we should used mocked responses to speed up testing for the data-client. But I still wanted to have tests for the actual client to ensure that queries are being sent correctly, and that the functions for setting schema/dropping data work.

src/GraphQLClient/config/opendialog-graphql.php
```diff
@@ -0,0 +1,154 @@
...
+type Scenario  {
+    id: ID!
+    od_id: String! @search(by:[hash])
+    name: String! @search(by:[hash])
```
bendeeson, line 15: This is the GraphQL schema to be passed to DGraph. It only contains types. DGraph generates the rest of the schema including data types etc from it.

src/GraphQLClient/GraphQLClientServiceProvider.php
```diff
@@ -0,0 +1,60 @@
...
+            $baseUrl = config('opendialog.graphql.DGRAPH_BASE_URL');
+            $port = config('opendialog.graphql.DGRAPH_PORT');
+            if (!$baseUrl) {
+                throw new GraphQLClientConfigException("opendialog.graphql.DGRAPH_BASE_URL config value is missing.");
+            }
+            if (!$port) {
+                throw new GraphQLClientConfigException("opendialog.graphql.DGRAPH_PORT config value is missing.");
+            }
+
+            $instanceType = config('opendialog.graphql.DGRAPH_INSTANCE_TYPE');
```
patshone-gsl, line 31: I wonder whether this logic might be better living outside of the service provider. I don't really know what the idiomatic laravel way of doing this would be - perhaps a client factory?

src/GraphQLClient/Tests/SlashGraphQLEndpointTest.php
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public function getPackageProviders($app)
+    {
+        return [
+            GraphQLClientServiceProvider::class
+        ];
+    }
+
+//    /**
```
patshone-gsl, line 29: Does the whole test need to be commented out here? - is the @group skip annotation not enough

src/GraphQLClient/config/opendialog-graphql.php
```diff
@@ -0,0 +1,152 @@
+<?php
+
+
+return [
+    'DGRAPH_INSTANCE_TYPE' => env('DGRAPH_INSTANCE_TYPE', null),
+    'DGRAPH_BASE_URL' => env('DGRAPH_BASE_URL', null),
+    'DGRAPH_PORT' => env('DGRAPH_PORT', null),
+    'DGRAPH_AUTH_TOKEN' => env('DGRAPH_AUTH_TOKEN', null),
+    'SLASH_GRAPHQL_API_KEY' => env('SLASH_GRAPHQL_API_KEY', null),
+    "schema" => <<<'GQL'
```
patshone-gsl, line 10: It feels like it would be better for this schema to live in a separate file rather than the config object. I'd be happier diffing if the changes to the schema itself were separate from changes to the config

patshone-gsl (changes requested):
We should fix the commented test and move the schema into a separate file, but then I think this is good to go

src/GraphQLClient/config/opendialog-graphql.php
bendeeson: Makes sense. I'll factor the GraphQL into a separate file and load with `file_get_contents()`

patshone-gsl (approved):
Looks good

PR #450, opened 2021-03-02
Incoming-outgoing intent matching: Basic end-to-end & IntentContext

elliotmassen:
This PR begins the work necessary to build out incoming-outgoing intent matching in the conversation engine, and covers a basic end-to-end case, as well as use of the intent context. ~A temporary base branch has been created to make reviewing easier, as this PR is dependent on two currently unmerged PRs: #446 & #449.~

The basic end-to-end case covers a non-ongoing conversation in which an incoming (user) intent is matched as a turn request, and an outgoing (app) intent is matched as a turn response. This has entailed changes to some existing code and end-to-end test (in `CoreTest`) that presumed that the conversation engine would explicitly set the outgoing intent in case of a no-match, as opposed to explicitly setting the incoming intent and matching the corresponding outgoing intent via regular means.

It's worth noting that this PR works on an understanding of the following as distinct concepts, which give us two dimensions on which to consider intents (and therefore four distinct combinations of intent types):
 - **Incoming/outgoing intents**: These mirror the HTTP request/response cycle, in that we receive an incoming (utterance which we interpret to an intent) and return an outgoing. These also map directly to user/app intents for now, ie. a user only ever gives an incoming intent, the app only ever gives an outgoing intent.
 - **Request/response intents**: These are not tied to the incoming-user/outgoing-app binary, and instead denote the position on an intent within a turn. Each turn has a set of request intents and a set of response intents. An incoming intent can be either a request or response intent, as can an outgoing intent.

This PR has updated some variable and method names to better reflect these distinctions, as it will become increasingly important to recognise whether an intent is incoming or outgoing as separate from recognising whether an intent is request or response.


src/ConversationEngine/ConversationEngine.php
```diff
@@ -64,112 +64,84 @@ public function getNextIntents(UtteranceAttribute $utterance): IntentCollection
...
             $intent->getTurn()->getODId()
         );
         ContextService::saveAttribute(
-            self::CONVERSATION_CONTEXT.'.'.Intent::CURRENT_INTENT,
+            $conversationContextId .'.'.Intent::CURRENT_INTENT,
             $intent->getODId()
         );
         ContextService::saveAttribute(
-            self::CONVERSATION_CONTEXT.'.'.Intent::CURRENT_SPEAKER,
+            $conversationContextId .'.'.Intent::CURRENT_SPEAKER,
```
istos, line 143: We should also be creating a UserHistoryRecord here - something to discuss about where is the most appropriate place to do this.

src/ConversationEngine/Reasoners/ConditionFilter.php
```diff
@@ -4,22 +4,40 @@
...
+     * @param bool $populateIntentContext
      * @return ODObjectCollection
      */
-    public static function filterObjects(ODObjectCollection $objects): ODObjectCollection
+    public static function filterObjects(ODObjectCollection $objects, bool $populateIntentContext = false): ODObjectCollection
     {
-        return $objects->filter(function (ConversationObject $object) {
-            return self::checkConditionsForObject($object);
+        /** @var IntentContext|null $intentContext */
+        $intentContext = $populateIntentContext ? ContextService::getContext(IntentContext::getComponentId()) : null;
```
istos, line 26: I am a bit confused by this - I did not expect to find a reference to `intentContext` in a `filterObjects` method of the Condition Filter. Could do with some documentation that provides the background of why this is here.

src/ConversationEngine/Util/MatcherUtil.php
```diff
@@ -0,0 +1,56 @@
...
+
+use OpenDialogAi\ContextEngine\Contexts\BaseContexts\ConversationContext;
+use OpenDialogAi\ContextEngine\Facades\ContextService;
+use OpenDialogAi\Core\Conversation\Conversation;
+use OpenDialogAi\Core\Conversation\Intent;
+use OpenDialogAi\Core\Conversation\Scenario;
+use OpenDialogAi\Core\Conversation\Scene;
+use OpenDialogAi\Core\Conversation\Turn;
+
+class MatcherUtil
```
istos, line 15: If we called this `ConvContext` the code would look like `ConvContext::currentScenarioId()`, `ConvContext::currentConversationId` which is more readable?

istos (changes requested):
Only having it as Request changes for that comment on the ObjectFilter referring to intents.

PR #451, opened 2021-03-03
Incoming-outgoing intent matching: Abstraction of matchers

elliotmassen:
This PR continues the work started in #450 expanding of the incoming & outgoing intent matcher classes (`IncomingIntentMatcher` & `OutgoingIntentMatcher` respectively). The incoming intent matching is (implicitly) parameterised by the user's utterance and the conversation context, while the outgoing intent matching (implicitly) parameterised by only the conversation context. This strong dependency on the conversation context opens space for future features such as transitions and virtual intents.

## Description

Both incoming & outgoing intents are handled different dependent on two key cases:

1. Whether the user is currently in a conversation.
2. Whether the user is within a turn.

In the first case, we handle the intent as a **request intent** and query for objects using the starting behaviour (as the user is not in a conversation). If the user is in a conversation, we consider the second case. If the user is not in a turn, we handle the intent as **request intent** and query for objects using the open behaviour, or turns with a matching valid origin. If the user is in a turn, we handle the intent as a **response intent** and select from the available response intents of the current turn.

## Testing

This PR includes two types of tests. There are tests which directly test the various cases available in each matcher class. These tests mock the various selectors, set the initial conversation context state and essentially test that under these conditions that the matchers identify the correct case (eg. is the user in a conversation, are they mid-turn). The other type of tests are similar but from one-step up. They test that the conversation engine's `updateState` method is capable of manipulating the conversation context between intent matches. The tests defined here begin to test the broader cases that a user may be in such as: a test for matching an incoming starting-request intent and then matching its related outgoing response intent. The reason this broader testcase was added was to ensure that we have a way of testing that the conversation engine is managing the state as we expect. It will be worth expanding on these tests as we add features and change functionality to ensure that it captures the range of potential cases a user may be in.

## Functionality overview

This PR accounts for the following functionality:

 - Starting a conversation.
 - Handling the incoming-request outgoing-response pairing.
 - Handling the incoming-response outgoing-request pairing.
 - Moving between turns via the open behaviour.
 - Moving between turns via valid origins.
 - Completing a conversation.

## Clarification

There is a clarification needed around how scenes are moved between. This PR presumes that it is not possible to move from one scene to another. Of course, once we have transitions this will provide explicit means to move scenes, but there is a question over whether it is possible to move scene without one. An example would be the following psuedo-CDL snippet:

```yaml
 - scene1:
     behavior: [starting]
     turns:
         - turn1:
             behavior: [starting]
             request:
               - u: intent.core.welcome
             response:
               - u: intent.app.welcomeOutput
         - turn2:
             behavior: [open]
             request:
               - u: intent.core.continue
             response:
               - u: intent.app.continueOutput
 - scene2:
     behavior: [open]
     turns:
         - turn1:
             behavior: [starting]
             request:
               - u: intent.app.faq
             response:
               - u: intent.app.faqOutput
```

The question is: after receiving `intent.app.welcomeOutput`, without a transition, does the user's next utterance attempt to match `intent.app.continue`?

## Note on lack of abstraction

As this work is still in progress, no major abstraction or refactoring has yet taken place. This PR certainly has multiple places where this could take place (specifically in the matcher class, and their tests). It seems best to let the work fully play out, and the related tests develop, before embarking on any abstraction or refactoring, so as to ensure that the code doesn't unnecessarily tie itself in knots at this stage. Please see comments below acknowledging the places where such work would be possible.


src/ConversationEngine/Reasoners/IncomingIntentMatcher.php
```diff
@@ -4,30 +4,150 @@
...
+
+        $conversation = ConversationSelector::selectConversationById(
+            new ScenarioCollection([$scenario]),
+            MatcherUtil::currentConversationId()
+        );
+
+        $scene = SceneSelector::selectSceneById(
+            new ConversationCollection([$conversation]),
+            MatcherUtil::currentSceneId()
+        );
```
elliotmassen, line 73: This use of the selector classes is found in both the matcher classes and could be abstracted at a later point.

src/ConversationEngine/Tests/IncomingIntentMatcherTest.php
```diff
@@ -78,38 +141,110 @@ private function updateStateToUndefined()
             $conversationContextId .'.'.Intent::CURRENT_INTENT,
             Intent::UNDEFINED
         );
```
elliotmassen, line 143: This setup of the conversational state is found across multiple testcases and could be abstracted at a later point.

src/ConversationEngine/Tests/IncomingIntentMatcherTest.php
```diff
@@ -78,38 +141,110 @@ private function updateStateToUndefined()
...
         ConversationSelector::shouldReceive('selectStartingConversations')
             ->once()
             ->andReturn(new ConversationCollection([$conversation]));
 
         $scene = new Scene();
-        $scene->setODId('test_scene1');
+        $scene->setODId(self::TEST_SCENE_1);
         SceneSelector::shouldReceive('selectStartingScenes')
             ->once()
             ->andReturn(new SceneCollection([$scene]));
```
elliotmassen, line 244: This mocking of selectors is found in multiple testcases and could be abstracted at a later point.

src/ConversationEngine/Reasoners/OutgoingIntentMatcher.php
```diff
@@ -13,51 +13,154 @@
...
+            new SceneCollection([$scene]),
+            MatcherUtil::currentIntentId()
+        );
+
+        $turns = $openTurns->concat($turnsWithMatchingValidOrigin);
+        $turns = $turns->unique(function (Turn $turn) {
+            return $turn->getODId();
+        });
+
+        $intents = IntentSelector::selectResponseIntents($turns, false);
```
istos, line 122: If I am looking for a request intent why are we selecting response intents here?

src/ConversationEngine/Reasoners/OutgoingIntentMatcher.php
```diff
@@ -13,51 +13,154 @@
...
-            $intents = IntentSelector::selectResponseIntents(new TurnCollection([$turn]), false);
-
-            if ($intents->isEmpty()) {
-                throw new NoMatchingIntentsException();
+            if (MatcherUtil::currentConversationId() == Conversation::UNDEFINED) {
+                // Its a non-ongoing conversation
+                return self::asStartingRequestIntent();
+            } else {
+                // Its an ongoing conversation
+                if (MatcherUtil::currentIntentIsRequest()) {
```
istos, line 43: There is something in the naming of all of this that confuses me. I will take some notes in terms of writing out my understanding of what is going on here and we can discuss. 

So, we are looking to get a single matching outgoing intent. This means that we have processed the incoming intent, have matched it, and now need to select the outgoing intent starting from the "position" that the incoming intent placed us in. 

If the incoming intent (is that what the `currentIntentIsRequest` checks for?) was matched as a Request then we will potentially be looking for a response (unless a transition placed us somewhere else - which we are not handling here for now). So the code is saying 

"get me an outgoing intent `asResponseIntent` because the incoming one was a Request" 

or

"get me an outgoing intent `asRequestIntent` because the incoming one was a Response"

src/ConversationEngine/Reasoners/OutgoingIntentMatcher.php
```diff
@@ -13,51 +13,154 @@
...
+            true
+        );
+
+        $turn = TurnSelector::selectTurnById(
+            new SceneCollection([$scene]),
+            MatcherUtil::currentTurnId(),
+            true
+        );
+
+        $intents = IntentSelector::selectResponseIntents(new TurnCollection([$turn]), false);
```
istos, line 158: This feels right - I am looking for response intents and I am selecting response intents.

istos (changes requested):
I think this is almost there but good to walk through it together once we can.

src/ConversationEngine/Reasoners/IncomingIntentMatcher.php
```diff
@@ -4,30 +4,150 @@
...
+     * @throws EmptyCollectionException
+     */
+    private static function asResponseIntent(): Intent
+    {
+        $scenario = ScenarioSelector::selectScenarioById(MatcherUtil::currentScenarioId(), true);
+
+        $conversation = ConversationSelector::selectConversationById(
+            new ScenarioCollection([$scenario]),
+            MatcherUtil::currentConversationId()
+        );
```
elliotmassen, line 102: We can avoid this "tunnelling" down the conversation objects as our "currentSceneId" & "currentTurnId" (etc) will be UID's. We can confidently retrieve them from the graph without going thru the whole selector process - this is only necessary when evaluating the next possible intents (eg. when its a starting request, or end of a turn, or transition).

PR #452, opened 2021-03-12
Incoming-outgoing intent matching: No-matching

elliotmassen:
This PR implements functionality for when an incoming intent can not be matched. It's worth noting that ultimately if an incoming, or outgoing, intent can not be matched an empty intent collection will be returned (leaving us open to consider the best way to let the user know that something has gone wrong). In the case of incoming intents we have a desired process to attempt before this point however. This process breaks down as so.

**If the user is not in an ongoing conversation**:

1. Update the utterance's callback ID to `intent.core.NoMatch` and re-run incoming intent matching.

**If the user is in an ongoing conversation**:

1. Update the utterance's callback ID to `intent.core.TurnNoMatch` and re-run incoming intent matching.
2. Update the utterance's callback ID to `intent.core.SceneNoMatch` and re-run incoming intent matching.
3. Update the utterance's callback ID to `intent.core.ConversationNoMatch`, unset the user's current conversation and re-run incoming intent matching.
4. Update the utterance's callback ID to `intent.core.NoMatch`, unset the user's current conversation and re-run incoming intent matching.

### DIFF
--- a/src/ConversationEngine/Reasoners/IncomingIntentMatcher.php
+++ b/src/ConversationEngine/Reasoners/IncomingIntentMatcher.php
@@ -43,7 +43,7 @@ class IncomingIntentMatcher
     public static function matchIncomingIntent(): Intent
     {
         if (ConversationContextUtil::currentConversationUid() == Conversation::UNDEFINED) {
-            // It's a non-ongoing conversation
+            // It's not an ongoing conversation
             try {
                 return self::matchNextIntentAsStartingRequestIntent();
             } catch (EmptyCollectionException $e) {
--- a/src/ConversationEngine/Reasoners/IncomingIntentMatcher.php
+++ b/src/ConversationEngine/Reasoners/IncomingIntentMatcher.php
@@ -4,7 +4,10 @@
 namespace OpenDialogAi\ConversationEngine\Reasoners;
 
 
-use Illuminate\Support\Facades\Log;
+use OpenDialogAi\AttributeEngine\CoreAttributes\UtteranceAttribute;
+use OpenDialogAi\ContextEngine\Contexts\BaseContexts\ConversationContext;
+use OpenDialogAi\ContextEngine\Contexts\User\UserContext;
+use OpenDialogAi\ContextEngine\Facades\ContextService;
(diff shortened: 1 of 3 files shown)


src/ConversationEngine/Reasoners/IncomingIntentMatcher.php
```diff
@@ -23,6 +27,11 @@
 
 class IncomingIntentMatcher
 {
+    const GLOBAL_NO_MATCH_INTENT_ID = 'intent.core.NoMatch';
```
istos, line 30: Is this a `SCENARIO_NO_MATCH` ?

src/ConversationEngine/Reasoners/IncomingIntentMatcher.php
```diff
@@ -33,11 +42,17 @@ class IncomingIntentMatcher
      */
     public static function matchIncomingIntent(): Intent
     {
-        try {
-            if (ConversationContextUtil::currentConversationUid() == Conversation::UNDEFINED) {
-                // It's a non-ongoing conversation
+        if (ConversationContextUtil::currentConversationUid() == Conversation::UNDEFINED) {
+            // It's a non-ongoing conversation
```
istos, line 46: `// It's not an ongoing conversation`  <- this would be a clearer way of saying it, maybe?

istos (changes requested):
A couple of small comments - love how concise all this can be 👍

src/ConversationEngine/Reasoners/IncomingIntentMatcher.php
elliotmassen: You could say that, I decided against it as at this point it represents a literal no-match (whereas you could say the others represent a no-match within a specific scope). I thought that naming it scenario no-match may confuse things as this is used for non-ongoing conversation requests too.

src/ConversationEngine/Reasoners/IncomingIntentMatcher.php
elliotmassen: Ha good point! I've now updated that.

istos (approved):

PR #463, opened 2021-04-08
Transitions

elliotmassen:
This PR implements transitions. Transitions allows the conversational state to be updated, therefore moving dialog to a different point in the applications (a different conversation, scene or turn). This current implementation only implements transitions for moving conversation and scene, as turn transitions require a little more thinking to properly implement. This PR also implements the `listens` functionality, which allows an intent to specify a specific previous intent that is "linked" to. `listens` gives intents listening for the correct previous intent priority over other matched intents, **but it does not supercede interpretation**.


src/ConversationEngine/Util/ConversationalState.php
```diff
@@ -44,15 +44,24 @@ public static function createStateFromIntent(Intent $intent): self
...
+        $intentIsRequest = $intent->isRequestIntent();
 
-        if ($intent->getBehaviors()->hasBehavior(Behavior::COMPLETING_BEHAVIOR)) {
+        $transition = $intent->getTransition();
+
+        if (!is_null($transition)) {
+            $conversationId = $transition->getConversation() ?? Conversation::UNDEFINED;
+            $sceneId = $transition->getScene() ?? Scene::UNDEFINED;
+            $turnId = $transition->getTurn() ?? Turn::UNDEFINED;
+            $intentIsRequest = false;
```
elliotmassen, line 55: This is the crux of the transitions implementation!

src/ConversationEngine/Reasoners/ListensFilter.php
```diff
@@ -0,0 +1,40 @@
...
+     * based on the previous intent. If the intent's are listening are listening for the correct intent, only those
+     * intents will be returned, otherwise only intents that weren't listening for anything in specific will be returned.
+     * This filter is intended for use after interpretation, as the `listens` directive is not a replacement for
+     * interpretation.
+     *
+     * @param IntentCollection $intents
+     * @param string $intentId
+     * @return IntentCollection
+     */
+    public static function filter(IntentCollection $intents, string $intentId): IntentCollection
```
elliotmassen, line 23: This is where `listens` are handled.

src/ConversationEngine/Util/ConversationalState.php
```diff
@@ -44,15 +44,24 @@ public static function createStateFromIntent(Intent $intent): self
         $sceneId = $intent->getScene()->getUid();
         $turnId = $intent->getTurn()->getUid();
         $intentId = $intent->getUid();
+        $intentIsRequest = $intent->isRequestIntent();
 
-        if ($intent->getBehaviors()->hasBehavior(Behavior::COMPLETING_BEHAVIOR)) {
+        $transition = $intent->getTransition();
+
+        if (!is_null($transition)) {
```
patshone-gsl, line 51: Is there a reason why this isn't just

```
 if ($transition) {
```

patshone-gsl (approved):

PR #464, opened 2021-04-09
Feature/opndlg 454 conditions

stu-greenshoots:
This PR fills out the missing pieces for Conditions in core:
+ Update the condition normaliser to be able to serialize and deserialise conditions
+ Updates the conversation data client to return the required data for conditions on all conversation objects
+ Updates the conversation object normaliser to take conditions into account
+ Updates the graphql schema to allow for the required condition fields.

Note - this PR removes the union type for operation parameters and just accepts a string type for values. This is probably something to review 

Conditions will be passed into and returned from the API in the following format:
```
    "conditions": [
        {
            "operation": "eq",
            "operationAttributes": [
                {
                    "id": "attribute",
                     "value": "user.age"
                }
                ],
            "parameters": [{ "id": "value", "value": "18" }]
        }
    ],
    ```

### CODE REVIEW

File: src/Conversation/DataClients/ConversationDataClient.php [hunk 1]
```diff
@@ -143,7 +170,16 @@ public function query()
             interpreter
             behaviors
             conditions {
```
patshone-gsl, line 172: I can't figure out how to reference this properly, but I guess we should add the conditions graphql to the FULL_SCENE_FRAGMENT as well

PR #467, opened 2021-04-13
Fix/remove conversation array before set

stu-greenshoots:
This is in draft state as something to discuss. 

I have not been able to get a single query to work that removes any currently set data for behaviours and conditions, and then sets the new values. 

So this crudely performs another query to set the remove the values before the update query. 

If this is the solution we go for, we will need to think about:
- the atomosity of the calls - currently 2 separate queries with no error handling
- We don't check the input fields (which works at the moment only because we know the front end will send all fields). For this to be better, it should only clear down array fields that are also in the update request


stu-greenshoots:
I am merging this as is knowing it needs discussion going into next sprint

PR #473, opened 2021-04-22
Expected attributes

elliotmassen:
This PR implements expected attributes for intents in the conversation engine. When an intent is interpreted it may produce attributes (entities in some NLU providers), which are attached to the interpreted intent. Expected attributes allows the conversation designer to specify which contexts these attributes should be saved to.

For example, an intent may have the following:

 - Interpreted attributes: `first_name`, `last_name`, `age`, `email`
 - Expected attributes: `user.first_name`, `user.last_name`, `user.age`

This would mean that we want `first_name`, `last_name` & `age` attributes to be saved in the `user` context. `email` is not accounted for in the expected attributes and will therefore be put in the `session` context by default.


src/ConversationEngine/Reasoners/ExpectedAttributeHandler.php
```diff
@@ -0,0 +1,83 @@
...
+
+        $attributeToContextMap = $unexpectedAttributesMap->merge($expectedAttributesMap);
+
+        $updatedContexts = new Set();
+        foreach ($attributeToContextMap as $attributeName => $contextName) {
+            $updatedContexts->add($contextName);
+            ContextService::saveAttribute($contextName.'.'.$attributeName, $interpretedAttributes->get($attributeName));
+        }
+
+        // todo: ContextService could keep track of updated contexts throughout the session and only persist once at the end
```
patshone-gsl, line 48: Let's remove this todo and create a ticket instead

patshone-gsl (approved):
Looks good

PR #476, opened 2021-04-26
Updates the header name used to secure communication with Slash DGraph

patshone-gsl:
This PR updates the auth header used with api key to secure communication with slash dgraph instance.

Somewhat confusingly the documentation does state in the overview of DGraph authentication - https://dgraph.io/docs/cloud/admin/overview/#sidebar/ that the header should be `X-Auth-Token` however the example on this page - https://dgraph.io/docs/cloud/admin/authentication/ describes using `Dg-Auth` as the header name and it is this value that functions correctly.


elliotmassen (commented):
Yeah that documentation is definitely ambiguous! One reading of it sounds like the `Dg-auth` header is meant specifically for `/admin` endpoints - I'm wondering if we might need to include both? However if only having the `Dg-auth` header works for `/alter` & `/query` then no need!

PR #478, opened 2021-04-28
Core interpreters

elliotmassen:
This PR adds a core interpreter for Dialogflow and updates the LUIS interpreter to use the V3 API. This PR also removes other interpreters from the default interpreter registration. There is certainly some cleanup to be done regarding interpreter configuration, however this seems better handled in the upcoming work that will require configuration to be editable through the admin UI.


src/InterpreterEngine/Dialogflow/DialogflowClient.php
```diff
@@ -0,0 +1,120 @@
...
+    public function sendRequest($message, $projectId = null, $environment = null): AbstractNLUCustomRequest
+    {
+        $projectId = $projectId ?? $this->getDefaultProjectId();
+
+        /** @var UserAttribute $userAttribute */
+        $userAttribute = ContextService::getAttribute(UtteranceAttribute::UTTERANCE_USER, UserContext::USER_CONTEXT);
+
+        $sessionId = $userAttribute->getUserId() ?: uniqid();
+
+        $client = new \GuzzleHttp\Client();
```
patshone-gsl, line 96: Is there a reason why the client is created afresh for each request?

src/InterpreterEngine/Luis/LuisClient.php
```diff
@@ -35,7 +25,7 @@ public function __construct(Client $client, $config)
         $this->client = $client;
         $this->appUrl = $config['app_url'];
         $this->appId = $config['app_id'];
-        $this->staging = $config['staging'] ? 'TRUE' : 'FALSE';
+        $this->slot = 'staging';
```
patshone-gsl, line 28: I wonder whether this should be set in config?

PR #481, opened 2021-05-04
Transitions & conditions import/export bug

elliotmassen:
This PR fixes some issues with the serialisers, mostly relating to passing context down. It also brings the import/export serializers closer to the regular ones.


stu-greenshoots (approved):
This looks good to me. I think it is missing the actions normalisers, but they are now present in this PR:
https://github.com/opendialogai/core/pull/480

patshone-gsl (approved):

PR #485, opened 2021-05-12
Feature/opndlg 650 response conversation engine updates

stu-greenshoots:
Adds new methods to the response engine to deal with fetching the message templates from the graph using the new MessageTemplateDataClient. Condition filtering is handled by the ConversationFilter class as per other conversational objects

For now, the new methods sit alongside the old methods - the next PR will be to clean these up.

The OpenDialogController now passes in the `uid` rather than `od_id` to get the message templates.

The conversation data client has been updated to save message templates along with intents on saving of a full scenario tree.


src/ResponseEngine/Service/ResponseEngineService.php
```diff
@@ -189,6 +236,21 @@ public function buildTextFormatterErrorMessage(string $platform, string $message
...
+     * @param MessageTemplateCollection $messageTemplates
+     * @return GraphMessageTemplate|null
+     */
+    private function selectGraphMessageFromConditions(MessageTemplateCollection $messageTemplates): ?GraphMessageTemplate
+    {
+        $passing = ConditionFilter::filterObjects($messageTemplates);
+
+        return $passing->sortByDesc(function (GraphMessageTemplate $messageTemplate) {
+            return $messageTemplate->getConditions()->count();
+        })->first();
```
elliotmassen, line 251: We should make sure this behaviour is documented somewhere, as it's different to how other conditions compete, and not immediately obvious.

elliotmassen (approved):

src/ResponseEngine/Service/ResponseEngineService.php
stu-greenshoots: Yeah, good point. I just checked previous message docs and this logic isn't mentioned in there. I'll put a note on this ticket to make sure it gets picked up in the new docs

PR #486, opened 2021-05-13
Interpreter configuration

elliotmassen:
This PR introduces the concept of the `ComponentConfiguration`, which provides the ability to store configuration for a component in the database. This is used concretely to populate `InterpreterConfiguration` objects. Each interpreter provides it's own subclass of `InterpreterConfiguration`, which populates its properties from the given configuration array, and more importantly can specify which fields are required for the interpreter to function as expected.

This PR lays the groundwork for API's to be created for management of the configurations, a subsequent PR will be published that covers the conversation engine integration of this work.


src/InterpreterEngine/Configuration/InterpreterConfiguration.php
```diff
@@ -0,0 +1,32 @@
...
+
+
+namespace OpenDialogAi\Core\InterpreterEngine\Configuration;
+
+
+use OpenDialogAi\Core\InterpreterEngine\Exceptions\InvalidConfigurationDataException;
+
+interface InterpreterConfiguration
+{
+    public function __construct(string $name, array $data);
```
stu-greenshoots, line 11: I'm not sure we want a constructor in an interface in this way as it removes the flexibility of an implementation. It's also discouraged by PHP:
https://www.php.net/manual/en/language.oop5.interfaces.php

Does this method have overlap with the static create method?

src/InterpreterEngine/Dialogflow/DialogflowInterpreterConfiguration.php
```diff
@@ -0,0 +1,47 @@
...
+use Ds\Map;
+use OpenDialogAi\Core\InterpreterEngine\Configuration\InterpreterWithEntitiesConfiguration;
+
+class DialogflowInterpreterConfiguration extends InterpreterWithEntitiesConfiguration
+{
+    const APP_URL = 'app_url';
+    const INTENTS = 'intents';
+
+    protected static array $required = [
+        self::APP_URL,
```
stu-greenshoots, line 16: How is the `KEY` used in this configuration? I would assume it needs one similar to LUIS / QnA

stu-greenshoots (changes requested):
This looks great to me, really like the implementation.

I have left a couple of comments - with the Dialogflow set up, I probably just don't understand how that works!

src/InterpreterEngine/Configuration/InterpreterConfiguration.php
elliotmassen: Ah yes, good point - I think I can do away with the signature here and keep the method in the abstract class for re-use.

src/InterpreterEngine/Dialogflow/DialogflowInterpreterConfiguration.php
elliotmassen: I'll be changing this specific configuration class shortly actually. The Dialogflow integration, that I brought into core recently, is in fact an integration for use with a proxy lambda function (which gave us support for Dialogflow Knowledge Bases - as the PHP library doesn't yet support them). The integration (& configuration) will need updating, which will in turn introduce the key here, good spot!

stu-greenshoots (approved):

PR #489, opened 2021-05-18
Lazy loading of attributes for custom contexts.

istos:
This moves the lazy loading of attributes for custom contexts into the AbstractCustomContext class so that all custom contexts can use it.


stu-greenshoots (approved):
Looks good to me. I wondered whether this meant that persistent contexts might be loaded multiple times, but the PersistedContext abstract class has a custom getAttribute method, so we are good 👍

elliotmassen (approved):

PR #494, opened 2021-05-26
Updating Dialogflow interpreter to use SDK

elliotmassen:
This PR updates the Dialogflow interpreter to use the PHP SDK, as opposed to an external proxy. It also makes use of our [fork](https://github.com/opendialogai/google-cloud-php-dialogflow) of the SDK to get support for Knowledge Bases (via v2beta1).


patshone-gsl (approved):
I think this looks good to me and once google have updated their php-dialogflow client, we can just remove the repo - I'm assuming here that we would continue to use the google-cloud-dialogflow client as opposed to the full google/cloud dep

PR #496, opened 2021-05-27
Action configuration preparation

elliotmassen:
This PR builds on the work related to interpreter configurations in #486 & #490 by fully generalising component configurations. It adds a `ComponentHelper` class which parses component ID's into a recognised component type, which allows API's to determine the correct way to handle various component types when processing configurations. A subsequent PR will handle implementation of action configurations in the conversation engine and various services.


stu-greenshoots (approved):
All looks good and working

PR #500, opened 2021-06-03
Action engine refactor for configurations

elliotmassen:
This PR continues the work from #496 by fully implementing configurable actions in the action engine. Following the pattern set out in #490, the action service has been split into three separate services: one container service for action **components**, one container service for action **configurations**, and a service for performing actions. The reflection API has also been updated to include data on action configurations.

## Webhook action

This PR also include a new core action component: the webhook actions. This action sends a HTTP POST request to a configurable URL, which will send any attributes passed into the action. The response from the URL can specify attributes to be returned by the action also.

### Request example

```json
{
  "input_data": {
    "attributes": {
      "first_name": "Elliot",
      "last_name": "Massen",
    }
  }
}
```

### Response example

```json
{
  "successful": true,
  "output_data": {
    "attributes": {
      "full_name": "Elliot Massen"
    }
  }
}
```


src/ActionEngine/Service/ActionService.php
```diff
@@ -0,0 +1,93 @@
...
+                    "Skipping action %s (%s) because some required attributes does not exist",
+                    $name,
+                    $componentId
+                )
+            );
+            return ActionResult::createFailedActionResult();
+        }
+
+        try {
+            $actionResult = $action->perform($actionInput);
```
stu-greenshoots, line 54: Lets get a debug level log in for successful attempts also 👍

src/ActionEngine/Actions/WebhookAction.php
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * @inheritDoc
+     */
+    public function perform(ActionInput $actionInput): ActionResult
+    {
+        $webhookRequest = $this->createRequest($actionInput);
+
+        try {
+            $webhookResponse = Http::post($this->configuration->getWebhookUrl(), $webhookRequest);
```
stu-greenshoots, line 40: Are we using the HTTP facade elsewhere? We should think about migrating from Guzzle throughout

elliotmassen: I believe this is our first use of it (I remembered it coming up recently and thought it was worth giving it a spin). I agree, `Http` has useful testing methods too which is a lot easier that create mock clients and passing them through.

PR #502, opened 2021-06-03
Increase callback interpreter flexibility

istos:
This PR makes the callback interpreter a bit more permissive in that we will successfully match against callbacks and the intent ID, using both the intent name and the intent ID as possible match sources. 

It also adds some additional logging on intent matching.


src/Conversation/Intent.php
```diff
@@ -178,13 +179,23 @@ public function checkForMatch(): bool
...
         foreach ($this->interpretedIntents as $intent) {
-            if (($intent->getODId() == $this->getODId()) && ($intent->getConfidence() >= $this->getConfidence())) {
+            Log::info(sprintf("Comparing %s with confidence %s to %s with confidence %s",
+                    $this->getODId(),
+                    $this->getConfidence(),
+                    $intent->getOdId(),
+                    $intent->getConfidence(),
+                )
+            );
+            if ((($intent->getODId() == $this->getODId()) || ($intent->getName() == $this->getName()))
```
elliotmassen, line 189: This condition raises some questions for me around the purpose of names, which are probably worth surfacing now. I think the inclusion of `|| ($intent->getName() == $this->getName())` here might be opening us up to some additional problems/confusion outside the scope of what this PR seeks to achieve. This change would mean that _all_ intent interpretation, beyond just the callback interpreter, can be matched via the intent name. At the moment, we automatically set the intent name based off the intent ID, so this isn't a field a designer will be aware of. Even despite that, I think that including this creates further ambiguity around the purpose of the intent ID and name for us. Up until now, my best explanation would've been that the intent ID is for matching purposes (as in this line originally), and that the intent name is for human-readable purposes (eg. front-end UI, etc). This additional condition here gives them similar purposes, whereas I think it might be best to keep them quite separate to reduce any ambiguity.

elliotmassen (commented):
Looks great - allowing the callback interpreter to be able to handle text is major 🚀   I have some concerns around the changes to intent matching however, which I've detailed below.

src/Conversation/Intent.php
```diff
@@ -178,13 +179,23 @@ public function checkForMatch(): bool
...
+                && ($intent->getConfidence() >= $this->getConfidence())) {
                 // @todo currently this will give us the last matching intent out of a number of intents as the "winner"
                 // so we may want to consider ranking within a single intent as well if we got multiple matches.
                 $this->interpretation = $intent;
+                Log::info("Intent comparison positive.");
```
stu-greenshoots, line 194: Is there any extra context we can add to the log messages to know which intent has been matched?

istos: The context comes from the previous log message.

src/Conversation/Intent.php
istos: I think you are right @elliotmassen definitely something to clarify here. Also realised we can get the same functionality I was after without the name comparison because the interpreter converts text to an id when it generates the intent to compare against.

elliotmassen (approved):

stu-greenshoots (approved):

PR #506, opened 2021-06-08
PHP 8 Support

patshone-gsl:
This PR enables core to be used by php 8 based applications by providing a 7.4 || 8.0 version constraint for php and by updating a symfony/yaml dependency.

In addition this PR provides the ability to publish migrations from core to a configurable location. This makes it easier to see and use migrations from core.


patshone-gsl:
> * I noticed that [opendialogai/opendialog](https://github.com/opendialogai/opendialog), doesn't have a PR for bumping the Composer requirement too - don't we need one there too to fully enable PHP 8 usage for an application?

Yes that's probably fair - at the moment I'm validating the function of core as a dependency in a php 8 application by using it in another application, a good next step would be to test opendialogai/opendialog running under both php 7.4 and php 8

> * Maybe something for the future, but it might be worth considering adding a separate CI job that builds and tests the repos against a PHP 8 image, so that we can be sure that there are no PHP 8 specific issues. Perhaps it could be filtered to only run on release branches or similar, so that we don't overburden regular commits with the extra CI running time.

This is a good idea - I'll create a ticket to update the circle config so that it tests in both contexts

stu-greenshoots (approved):

PR #507, opened 2021-06-08
Cleans out mysql based conversation and message classes, files and logic

stu-greenshoots:
This PR aims to clean up the outstanding bits of code from OD 0.7.x where conversations and message templates  were stored in MySQL.

Some tests have been left in place (where they are marked as skipped) so that the functionality can be added back

MySQL Tables Deleted:
- conversation
- outgoing intents
- message template
- activity log
- conversation state log

Modules Removed:
- ConversationBuilder

Model Classes Deleted:
- Conversation
- OutgoingIntent
- MessageTemplate
- Conversation state log

Commnands removed:
- export conversations
- import conversations
- read / store statuses (deleted calls from migrations that used them)

Dependencies Removed:
- Spatie Activity Log

Major Methods removed:
- getMessageForIntent from response enginge (replaced by getGraphMessageForIntent)
- Get message from conditions removed from response engine

Tests missing:

- Message logging test
- Attribute extraction tests have been removed
- Request logging tests (method left in CoreTest)
- Next intent detection test (class left in place)
- User Attribute tests


elliotmassen (approved):
All works as expected 👍 

I wonder whether this may be an opportunity to squash the migrations - but if that's a larger task then perhaps best not to hold up this getting merged.

patshone-gsl (approved):
Works for me locally

PR #513, opened 2021-06-10
Converting conversational state's intent_is_request to turn_status

elliotmassen:
This PR changes the `intent_is_request` boolean field in the conversational state to a `turn_status` string psuedo-enum. `true` has become `MID_TURN` which represents that the previous intent was a request and that we are now anticipating responses, and `false` has become `OUT_OF_TURN` which represents that we are outside of a turn and ready to consider new ones. This paves the way for an upcoming PR that will add a new `IN_TURN` status.

It's worth noting that we still retain the concept of the `intent_is_request` boolean on intents, as turn status is only relevant to conversation state, not intents themselves.


elliotmassen:
https://github.com/opendialogai/core/pull/509 will need to be merged into this PR, and request-only intent handling updated to use the new turn status.

istos (approved):
All checks out 👍

PR #514, opened 2021-06-10
Turn transitions

elliotmassen:
This PR implements turn transitions by adding a new `IN_TURN` turn status value. This new value tells intent matches to only consider request intents of the current turn (which will have been set by the transition).


istos (approved):
Love how concise and clear it is.  👍

istos (approved):

PR #516, opened 2021-06-15
Allowing interpreters to throw exceptions to caller

elliotmassen:
This PR allows interpreters to throw exceptions to their caller. In regular use, these will be caught be the interpreter service, and an empty intent collection returned. However this becomes useful when testing an interpreter configuration as we can get more fine-grain errors regarding the configuration.


patshone-gsl (approved):

stu-greenshoots (commented):
I wonder if it would be better to have a more generic interpret exception that all interpreters could throw rather than a more specific NLU exception

PR #518, opened 2021-06-17
Feature/opndlg 824 url validation

stu-greenshoots:
As part of a security review, we are now validating URLs that come in as part of the component configuration to make sure they look valid and point to public IPs.

The PR updates the interpreters and webhook action to use guzzles 'allow_redirects = false' option so that internal redirects can't take place.

There is a new pre-configured client bound to the interpreter engine service layer that is used by all interpreters that need it.

The webhook action uses Laravel's new HTTP facade, so is set up in place


patshone-gsl (approved):
As we've discussed, we'll want to get to consensus as to whether we use the Laravel proxy over guzzle or guzzle directly - for now this is fine

PR #524, opened 2021-06-23
Conversation object copying

elliotmassen:
This PR adds a `copy` method to conversation objects to provide full cloning functionality. Ideally this would have made use of the `__clone` magic method, however it was not possible to handle circular references between parent & child objects without﻿access to the original object.


src/Conversation/ConversationObject.php
```diff
@@ -311,4 +319,56 @@ public static function allFields()
...
+    {
+        $this->clone = $clone;
+    }
+
+    /**
+     * @return static
+     */
+    public function copy()
+    {
+        $clone = clone $this;
```
stu-greenshoots, line 343: I can't follow this flow - we create a clone, set it to `$this->clone` then at the end of the method setClone to null again?

src/Conversation/Tests/ConversationModelStructureTest.php
```diff
@@ -97,4 +97,111 @@ public function testIdenticalNamedItemsInCollection()
...
+            $this->assertCount($originalScenario->getConversations()->count(), $duplicateScenario->getConversations());
+
+            for ($j = 0; $j < $originalScenario->getConversations()->count(); $j++) {
+                /** @var Conversation $originalConversation */
+                $originalConversation = $originalScenario->getConversations()->get($j);
+
+                /** @var Conversation $duplicateConversation */
+                $duplicateConversation = $duplicateScenario->getConversations()->get($j);
+
+                $this->assertNotSame($originalConversation, $duplicateConversation);
```
stu-greenshoots, line 123: Is it worth adding in some assetSame on some fields like od_id, name etc to ensure that the data is populated in the copy?

stu-greenshoots (commented):
This looks good and is working - just 1 commend on the flow of the copy method and a small comment on the test

src/Conversation/ConversationObject.php
elliotmassen: You're right, it is a little confusing - I'll add a method doc comment to make it clearer!

This method returns the cloned object, so the use of `$this->clone` is only related to handling circular references between objects, not for actually storing the clone long-term. We require this short-term reference to the clone because of circular references between objects, but after the recursion if complete we can unset the reference. For example, we may ask to copy a scenario object which has child conversation objects, however when we begin recursively copying those child conversation objects, we will come across a reference back to the scenario. Setting `$this->clone` means that when we reach that circular reference we can simply return that _already_ cloned reference, rather than cloning a new one (which causes an infinite loop). The check for this happens here:

https://github.com/opendialogai/core/blob/606e590a0bea1b98eba66f1fef2f782e31ca2e80/src/Conversation/ConversationObject.php#L365-L369

src/Conversation/Tests/ConversationModelStructureTest.php
```diff
@@ -204,4 +211,17 @@ public function testCopy()
             }
         }
     }
+
+    /**
+     * @param ConversationObject $original
+     * @param ConversationObject $duplicate
+     */
+    public function assertObjectCloned(ConversationObject $original, ConversationObject $duplicate): void
```
stu-greenshoots, line 219: 👍

src/Conversation/ConversationObject.php
stu-greenshoots: right yep, that helps, thanks :)

stu-greenshoots (approved):

PR #528, opened 2021-06-28
Conversation, scene & turn copying

elliotmassen:
This PR continues the work started in #524 by extending the copying functionality to conversation, scene & turn objects. This PR also finally begins to break up the conversation data client into various smaller clients - although there is still a lot further this can go (however this was out of scope of the PR).


src/Conversation/DataClients/BaseDataClient.php
```diff
@@ -455,4 +469,359 @@ protected function getRemoveSet(array $fields): array
...
+
+    /**
+     * Store a new Scenario object
+     *
+     * @param Scenario $scenario
+     *
+     * @return Scenario
+     * @throws ExceptionInterface
+     */
+    public function addScenario(Scenario $scenario): Scenario
```
stu-greenshoots, line 481: Should these go into their various new data clients now? IE this could go in the scenario data client?

stu-greenshoots (commented):
This all looks great, and awesome to start splitting up the data client. Just a question about whether some of those methods belong in the base or in the individual clients

src/Conversation/DataClients/BaseDataClient.php
elliotmassen: I think eventually, yes. The problem is that there's quite a few places that presume that the `ConversationDataClient` has these methods - so I was wary of pulling on the thread too much and causing more changes than necessary for the scope of this work. I figured that best solution for now was to bring them into the `BaseDataClient` at least, to enable the creation of the split clients, and then we can later circle back for the larger clean up work.

stu-greenshoots (approved):

elliotmassen:
I accidentally merged this into the previously merged feature branch... this has been merged into `1.x` in https://github.com/opendialogai/core/commit/f6ec1f87eda14275f3596dfebdc740e6f9dcafa9.

PR #529, opened 2021-06-29
Fix/opndlg 868 remove character escaping

stu-greenshoots:
This PR addresses some issues around character encoding and escaping.

It removes a method in the response engine service that replace `&` with `&amp;` and `%` with `%%`

The `&` character should always come through as the HTML entity, otherwise the XML string will be invalid.

The escaping of the % character is no longer needed after an update to the set Text method of WebchatMessage. This was previously supposing that the message text sent to it could be in a format that `sprintf()` would understand, but this is not the case as the XML will always be a representation of the end message text, not a format.

On the back of this PR, messages will now have to be entered in the front end manually escaped (eg &amp; and &lt;) 

There will be a follow up ticket to make sure that any message blocks do the escaping for the user


elliotmassen (approved):
I've checked this out in my local application and it all works as expected - all works great 👍

PR #531, opened 2021-06-29
Intent copying

elliotmassen:
This PR adds an `IntentDataClient` as a continuation of the work in #528, which methods enable copying of intents and their child messages.


src/Conversation/DataClients/IntentDataClient.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+
+namespace OpenDialogAi\Core\Conversation\DataClients;
+
+
+use OpenDialogAi\Core\Conversation\Exceptions\ConversationObjectNotFoundException;
+use OpenDialogAi\Core\Conversation\Facades\MessageTemplateDataClient as MessageTemplateDataClientFacade;
```
stu-greenshoots, line 8: Not sure why we rename the Facade on import..?

src/Conversation/DataClients/TurnDataClient.php
```diff
@@ -5,7 +5,7 @@
 
 
 use OpenDialogAi\Core\Conversation\Exceptions\ConversationObjectNotFoundException;
-use OpenDialogAi\Core\Conversation\Facades\MessageTemplateDataClient as MessageTemplateDataClientFacade;
+use OpenDialogAi\Core\Conversation\Facades\IntentDataClient as IntentDataClientFacade;
```
stu-greenshoots, line 8: Again - do we have to rename?

stu-greenshoots (changes requested):
It's probably present in other places and been copied in - but could you look at not naming the imported facades

src/Conversation/DataClients/TurnDataClient.php
elliotmassen: I believe its because the namespace that this class is within includes `IntentDataClient` (and the other data client classes), which means that it's automatically available under that name - the only other option is to remove the imports and use the FQN when calling the methods of the facade, eg. `\OpenDialogAi\Core\Conversation\Facades\IntentDataClient::addFullIntentGraph()`.

src/Conversation/DataClients/IntentDataClient.php
elliotmassen: See https://github.com/opendialogai/core/pull/531#discussion_r662340838

stu-greenshoots (approved):

PR #532, opened 2021-06-30
URL click utterance fix

elliotmassen:
This PR fixes a bug with URL click utterances. As the webchat controller hadn't updated it's outdated utterance type, the URL click condition was failing and allowing the utterance to be interpreted. This PR resolves that issue, and also updates the `LinkClick` service to work with the new 1.x utterance style. It also introduces a more generalised way for sensors to "ignore" utterances via an `IgnoreUtteranceException` which can be throw to its controller.


src/SensorEngine/Sensors/WebchatSensor.php
```diff
@@ -124,7 +128,10 @@ public function interpret(Request $request) : UtteranceAttribute
                     );
                 }
 
-                return $utterance;
+                $linkClickService = resolve(LinkClickInterface::class);
```
stu-greenshoots, line 131: Great, this makes much more sense here 👍

stu-greenshoots (approved):
All looks good to me

PR #533, opened 2021-06-30
Outgoing starting intent matching bug

elliotmassen:
This PR fixes a bug in which outgoing starting intents were incorrectly treated as a response intent. This is noticeable in the updated simulator, at the scenario-level, where it is now possible to see both initial user & app intents. This meant that the welcome conversation's response intent is incorrectly included in the initial intent listing. It _could_ also naturally occur if an incoming intent had a completing behaviour and there was a valid starting conversation with an outgoing request intent.


stu-greenshoots (approved):
Makes sense 👍

PR #534, opened 2021-07-06
Attribute, context, action fixes

elliotmassen:
This PR makes some key fixes to how attributes are handled. It completes the changes made in #522, by ensuring that actions persist output attributes to the user context when no context is specified (previously it would only do so if an invalid context ID was given). It also adds `is_true` & `is_false` operations as our current UI presumes all condition parameters to be strings, which means we don't have a way of conditioning boolean attributes properly.


stu-greenshoots (approved):
Thanks Elliot - this is good to go from me

PR #536, opened 2021-07-08
First & last seen attributes

elliotmassen:
This PR adds three new attributes, which are stored in the user context:

 - `first_seen`: A unix timestamp representing the first time we encountered the user, set prior to intent or message matching.
 - `last_seen`: A unix timestamp representing the last time we encountered the user, set after intent or message matching.
 - `seconds_since_last_seen`: An integer representing the number of seconds since the last time we encountered the user, set prior to intent or message matching.

As we may want to use these attributes in conditions or actions during intent matching or message selection, this determines where in the code base they are set. The following is useful to understand there placement:

1. Request received
2. Utterance analysis, determine the user, set all three attribute if new user, calculate `seconds_since_last_seen` as difference between `last_seen` and now.
2. Intent matching & message selection, conditions & actions may make use of `last_seen` which will represent the _previous_ encounter with the user if there was one
3. Set new `last_seen` value for use in subsequent requests
4. Return response


src/ConversationEngine/Reasoners/UtteranceReasoner.php
```diff
@@ -103,6 +104,8 @@ private static function saveCustomUserData(UtteranceAttribute $utterance): void
      */
     private static function prepareUser(UtteranceAttribute $utterance, UserAttribute $currentUser): void
     {
+        $now = Carbon::now();
```
stu-greenshoots, line 107: We could potentially use LARAVEL_START constant that gets defined at the start of the request? So it would be the same number where ever used

stu-greenshoots: There's an example of it being used in the RequestLoggerMiddleware

src/ConversationEngine/Reasoners/UtteranceReasoner.php
```diff
@@ -143,9 +146,20 @@ private static function prepareUser(UtteranceAttribute $utterance, UserAttribute
...
             // There was a pre-existing user, load their persistent context attributes
             ContextService::setupPersistentContexts($utterance->getUserId());
         }
+
+        $lastSeenUnix = ContextService::getAttributeValue(UserAttribute::LAST_SEEN, UserContext::getComponentId());
+
+        $then = new Carbon($lastSeenUnix);
+        $sinceLastSeen = $then->diffInSeconds($now);
+
+        ContextService::saveAttribute(UserContext::getComponentId().".".UserAttribute::SECONDS_SINCE_LAST_SEEN, $sinceLastSeen);
```
stu-greenshoots, line 163: we set the last seen twice new users - I guess to make sure its closer to now.

I think we can use the LARAVEL_START time for setting last seen and first seen and the time comparison. 

That way we avoid any unwanted non-zero differences and have a fixed time to work from

stu-greenshoots (changes requested):
Awesome, this is super useful to have back in place. 

I have comment about using now() a marker as its used in various places so wont always be the same.

src/ConversationEngine/Reasoners/UtteranceReasoner.php
elliotmassen: It actually gets set here just in case there are any conditions/actions that rely on it for new users (otherwise it'd be null, as it gets set _after_ all that). Despite that, using `LARAVEL_START` is a really nice solution to the potential difference between the two - I'll make that change :)

stu-greenshoots (approved):

PR #539, opened 2021-07-15
Feature/intent transition queries

stu-greenshoots:
This pull request adds 3 new methods to the Intent Data Client that can be used to fetch any intents that have a transition referencing either the conversation, scene or turn (1 method for each).

These are needed so that the relationship can be checked when deleting a conversational object so that the user can be warned about the transition. 

If a conversational object is deleted, but a transition referencing it stays in place, the entire scenario is in an invalid state.

The schema has been updated to allow searching on the transition items


src/Conversation/DataClients/IntentDataClient.php
```diff
@@ -69,4 +68,127 @@ public function getFullIntentGraph(string $intentUid): Intent
...
+                    conversation {
+                      id
+                      od_id
+                      name
+                    }
+                  }
+                }
+                transition (filter: { conversation: {eq: $conversationId}}) {
+                  conversation
+                }
```
elliotmassen, line 104: Would it be useful to retrieve the full transition object here so as to be able to differentiate between conversation, scene & turn transitions? If so, I think you can use a [parameterised nested `@cascade`](https://dgraph.io/docs/graphql/queries/cascade/#parameterized-cascade) to allow scene & turn transitions to be returned. Something like:

```graphql
transition (filter: { conversation: {eq: $conversationId}}) @cascade(fields:["conversation"]) {
  conversation
  scene
  turn
}
```

I think that would allow for the `null` values of conversation & scene transitions to be returned, which would let you know what kind of transition is in use (eg. conversation transitions have `null` for scene & turn, scene transitions have `null` for turn`, turn transition have no `null` values`)

elliotmassen (approved):
This looks great! Just a thought below, but not necessarily blocking unless you think it's useful.

PR #542, opened 2021-07-22
Updating virtual_intents field to virtual_intent

elliotmassen:
This PR updates the unused, outdated virtual intent*s* field on intents to an appropriate virtual intent field. **This includes an update to the graph schema.**


src/Conversation/VirtualIntent.php
```diff
@@ -18,14 +20,28 @@ public function __construct(string $speaker, string $intentId)
         $this->intentId = $intentId;
     }
 
-    public function getSpeaker()
+    public static function createEmpty(): VirtualIntent
+    {
+        $virtualIntent = new self('', '');
```
stu-greenshoots, line 25: can we not just initialise with null in the `new` above?

stu-greenshoots (approved):
Just a small comment to have a look at

src/Conversation/VirtualIntent.php
elliotmassen: Good point - I was trying something different here (as opposed to the root taken with transitions). The idea is that it shouldn't be otherwise possible to create a virtual intent with no speaker or intent ID. So having a specific method that is used only when you need to represent an empty transition seemed a potential way to standardise that functionality. Happy to change if it's more confusing.

elliotmassen: Scratch that - one of the later PR's changes this to be exactly how transitions are done. It's not possible to have a null-able object (like transitions or virtual intents) unfortunately, as the the data clients aren't set up to remove null-ed fields, and we have ambiguity over where `null` means removed or not present. So now these fields are null-able and when they are null it means it's an empty virtual intent (ie. removed).

PR #543, opened 2021-07-22
Update conversation engine interface

elliotmassen:
This PR refactors the conversation engine's primary method to return message templates as opposed to intents. This is prior work to the imminent virtual intent work, and is necessary because virtual intents allow for multiple outgoing intents to be selected (chained by virtual intent(s)), and therefore we want to be able to evaluate message template conditions and fill attributes iteratively. Therefore the conversation engine requires control over this functionality.

This PR reduces the scope of the response engine service to handle only the _formatting_ of messages, which has always been distinct from message selection and/or filling. This means that we can rely on the conversation engine to supply message templates (with markup), that the response engine will format into the appropriate structure for the current platform. When further platform integrations are added, it may be worth further refactoring the response engine to mirror the container service patterns used in other engines.


src/ConversationEngine/ConversationEngineInterface.php
```diff
@@ -3,14 +3,16 @@
...
-     * Given an utterance attribute.
+     * Given an utterance attribute, match incoming & outgoing intents given the user's conversational state, process
+     * the intents and prepare the outgoing intent's message templates.
+     *
      * @param UtteranceAttribute $utterance
-     * @return IntentCollection
+     * @return MessageTemplateCollection
      */
-    public function getNextIntents(UtteranceAttribute $utterance): IntentCollection;
+    public function run(UtteranceAttribute $utterance): MessageTemplateCollection;
```
elliotmassen, line 17: This change is the crux of this PR.

Tests/Feature/CrossConversationVirtualIntentsTest.php
```diff
@@ -23,15 +23,15 @@ public function testUVirtualBetweenConversations()
         $userContext = ContextService::createUserContext($startUtterance);
 
         $conversationEngine = resolve(ConversationEngineInterface::class);
-        $intents = $conversationEngine->getNextIntents($userContext, $startUtterance);
+        $intents = $conversationEngine->run($userContext, $startUtterance);
```
stu-greenshoots, line 26: Are these now message templates? The test update is a bit confusing as they are still named intents and seem to check for the same ids

src/ConversationEngine/Reasoners/MessageTemplateMatcher.php
```diff
@@ -0,0 +1,162 @@
...
+use OpenDialogAi\ConversationEngine\Exceptions\NoMatchingMessagesException;
+use OpenDialogAi\Core\Conversation\Intent;
+use OpenDialogAi\Core\Conversation\IntentCollection;
+use OpenDialogAi\Core\Conversation\MessageTemplate;
+use OpenDialogAi\Core\Conversation\MessageTemplateCollection;
+use OpenDialogAi\MessageBuilder\MessageMarkUpGenerator;
+
+class MessageTemplateMatcher
+{
+    const NO_INTENTS_ERROR_MESSAGE = 'No intents were matched, make sure that you have a scenario activated'
```
stu-greenshoots, line 23: In theory, we could have this interspersed between messages with content following multiple u_virtuals?

src/ResponseEngine/Message/Webchat/WebChatMessages.php
```diff
@@ -43,4 +43,23 @@ public function getMessageToPost(): array
 
         return $messagesToPost;
     }
+
+    /**
+     * @inheritDoc
+     */
+    public function merge(OpenDialogMessages ...$messages): OpenDialogMessages
```
stu-greenshoots, line 50: I was going to say we should pull this to the parent, but there isn't one. When we think about other platforms, we should make this method shared

stu-greenshoots (approved):

src/ConversationEngine/Reasoners/MessageTemplateMatcher.php
elliotmassen: That shouldn't be possible for it to be between messages, as this string is used for an artificially created message template (via `createErrorMessageTemplate`) which wouldn't have an intent, or associated virtual intent. It _could_ be the final message however.

Tests/Feature/CrossConversationVirtualIntentsTest.php
elliotmassen: Ah yes, you're right! I think this was done by the IDE's auto renamer! This test is actually one that it skipped as it's from 0.7 - it's removed in the subsequent PR.

src/ConversationEngine/Reasoners/MessageTemplateMatcher.php
stu-greenshoots: good point :+1:

PR #544, opened 2021-07-26
Virtual intents

elliotmassen:
This PR implements virtual intents. Virtual intents allow an outgoing intent to specify that after matching, a "virtual" incoming intent should be run through the conversation engine, as if the user had uttered it. Virtual intents are matched by intent ID, but they are not interpreted. They are still subject to conditions however.

To enable this functionality the the following were necessary:

 - Addition of a virtual utterance type: to allow the conversation engine and simulator to handle virtual utterances differently from regular ones.
 - Addition of the `VirtualIntentFilter` class: to filter intent collections using a virtual utterance. Eg. filter out all intents that don't match the virtual intent ID.
 - Update of `getNextIntents` to `getNextIntent`: this method previously returned an intent collection in anticipation of virtual intents, however the implementation has been outside of the method and therefore we can now return to presuming the method returns a single outgoing intent.
 - Ability to set filter callbacks on selectors: the simulator requires that, broadly, no filtering of conversation objects happens (ie. we don't want condition filtering, interpretation, etc). However we _do_ want to be able to filter via a virtual utterance. Therefore we now have the ability to set specific filters for selectors, which is used by the simulator to set the `VirtualIntentFilter` on the intent selector.


stu-greenshoots (approved):
Thanks @elliotmassen - this is all looking great. I have pulled in locally, stepped through the code and tested and prodded. I think this is good to merge and go into testing. Awesome

PR #545, opened 2021-07-29
OpenDialog interpreter

elliotmassen:
This PR renames the Callback interpreter to OpenDialog interpreter and introduces optional text similarity evaluation to the interpreter. To account for this, this PR updates the `configurations:create` command to update the default configuration to use the new name and the renamed component ID.

It also updates the interpreter service to pass the current intent being interpreted to interpreters through optional before/after callbacks. This is used by the OpenDialog interpreter to get the current intent's sample utterance and compare it to the provided text utterance from the user.

It also adds the ability for an interpreters result to specify that it should not be cached. This is used by the OpenDialog interpreter, as its use of the current intent (as specified above) in text similarity evaluation means that the interpretation is no longer idempotent in that instance, and therefore caching is not sensible. Although related to the new callback functions, this functionality is separated as it will not always be the case that an interpreter that is making use of the callback functions will not be idempotent (it may be simply logging something for example). It is especially useful for it to be decoupled even in the OpenDialog interpreter as it means that caching can be skipped only when text similarity evaluation has occurred, but not when it is disabled, or the utterance had a callback.


src/Console/Commands/CreateCoreConfigurations.php
```diff
@@ -39,22 +46,56 @@ public function handle()
...
+
+                    $this->info('Configuration updated.');
+                    return 1;
+                } catch (Exception $e) {
+                    $this->error(sprintf('Update failed, configuration was not updated: %s', $e->getMessage()));
+                    return 1;
+                }
+            } else {
+                $this->info('Okay, not updating.');
+                return 1;
```
stu-greenshoots, line 79: Also here?

src/Console/Commands/CreateCoreConfigurations.php
```diff
@@ -39,22 +46,56 @@ public function handle()
...
+            }
+        } else if ($configurationService->has(self::OPENDIALOG_INTERPRETER)) {
             $this->warn(sprintf(
                 "Configuration '%s' was not created, a configuration with that name is already registered.",
-                self::DEFAULT_CALLBACK
+                self::OPENDIALOG_INTERPRETER
             ));
-        } else if (!$componentService->has(CallbackInterpreter::getComponentId())) {
+        } else if (!$componentService->has(OpenDialogInterpreter::getComponentId())) {
             $this->warn(sprintf(
```
stu-greenshoots, line 87: we might want to return with a 1 here though - as this is unexpected

src/Conversation/DataClients/ConversationObjectDataClient.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+
+namespace OpenDialogAi\Core\Conversation\DataClients;
+
+
+class ConversationObjectDataClient extends BaseDataClient
```
stu-greenshoots, line 7: Questing whether we need another data client, but there actually isn't anywhere else this could live sensibly

src/InterpreterEngine/Interpreters/CallbackInterpreter.php
```diff
@@ -2,143 +2,12 @@
...
-use OpenDialogAi\Core\Conversation\Intent;
-use OpenDialogAi\Core\Conversation\IntentCollection;
-use OpenDialogAi\Core\InterpreterEngine\Callback\CallbackInterpreterConfiguration;
-use OpenDialogAi\InterpreterEngine\BaseInterpreter;
-
-class CallbackInterpreter extends BaseInterpreter
+/**
+ * @deprecated
+ */
+class CallbackInterpreter extends OpenDialogInterpreter
```
stu-greenshoots, line 8: Is this deprecated rather than just renamed as it could be referenced in someone's config?

src/InterpreterEngine/BaseInterpreter.php
```diff
@@ -20,9 +20,54 @@ abstract class BaseInterpreter implements InterpreterInterface, OpenDialogCompon
...
+     */
+    public function getBeforeInterpretationCallback(): ?callable
+    {
+        return $this->beforeInterpretationCallback;
+    }
+
+    /**
+     * @param callable $callback
+     */
+    public function setAfterInterpretationCallback(callable $callback): void
```
stu-greenshoots, line 60: specifying the type like this would prevent anything not callable being set, and negates having to check that is callable with `is_callable` in the relevant methods in the interpreter service - is that right?

src/InterpreterEngine/InterpreterEngineServiceProvider.php
```diff
@@ -22,7 +22,8 @@
 class InterpreterEngineServiceProvider extends ServiceProvider
 {
     const CORE_INTERPRETER_CONFIGURATIONS = [
-        CreateCoreConfigurations::DEFAULT_CALLBACK
+        CreateCoreConfigurations::DEFAULT_CALLBACK,
```
stu-greenshoots, line 25: Why do we want to add the new one rather than replace?

src/InterpreterEngine/Interpreters/OpenDialogInterpreter.php
```diff
@@ -0,0 +1,204 @@
...
+
+            /** @var OpenDialogInterpreterConfiguration $configuration */
+            $configuration = $this->getConfiguration();
+
+            if ($this->hasCurrentIntent() && $configuration->enableSimilarityEvaluation()) {
+                // Text similarity interpretation
+                $utteranceText = $utterance->getText();
+                $sampleUtterance = $this->getCurrentIntent()->getSampleUtterance();
+
+                if (!is_null($utteranceText) && !is_null($sampleUtterance)) {
```
stu-greenshoots, line 84: Maybe log a warning (or at least debug) in an else - we might if this has happened

src/InterpreterEngine/Service/InterpreterService.php
```diff
@@ -33,18 +43,23 @@ public function interpret(string $name, UtteranceAttribute $utterance): IntentCo
             $componentId = $interpreter::getComponentId();
         } catch (ConfigurationNotRegistered $e) {
             Log::error(sprintf("Interpreter %s is not a registered interpreter.", $name));
+            $this->intent = null;
```
stu-greenshoots, line 46: I guess we need to unset the intent each time in case its passed into the next interpreter? Is there a way to make sure its happened or do in a central way?

src/InterpreterEngine/Service/InterpreterService.php
```diff
@@ -143,4 +164,28 @@ public function setDefaultInterpreter(string $defaultInterpreterName): void
...
+        if (is_null($callback)) {
+            return;
+        }
+
+        $callback($this->intent);
+    }
+
+    public function afterInterpretation(BaseInterpreter $interpreter): void
+    {
+        $callback = $interpreter->getAfterInterpretationCallback();
```
stu-greenshoots, line 181: should we double check if callable wrap in a try/catch to try to handle anything unexpected?

src/InterpreterEngine/Similarity/SimilarityService.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        return collect($expected)
+            ->reduce(
+                fn ($carry, $e) => max($carry, $this->evaluate($e, $actual)),
+                0
+            );
+    }
+
+    /**
+     * Calculates the Levenshtein distance between two strings and returns a percentage value
```
stu-greenshoots, line 30: does it? I thought that was the levenshtein function

src/InterpreterEngine/Similarity/SimilarityServiceInterface.php
```diff
@@ -0,0 +1,32 @@
...
+     * Evaluates the similarity of the actual string from the expected strings.
+     * Eg.
+     *  (['hello', 'something'], 'world') => low score
+     *  (['hello', 'something'], 'hola') => high score
+     *
+     * @param array|string[] $expected
+     * @param string $actual
+     * @return float Between 0 and 1.
+     */
+    public function evaluateHighest(array $expected, string $actual): float;
```
stu-greenshoots, line 31: is this in preparation for multiple sample utterances?

src/InterpreterEngine/config/opendialog-interpreterengine.php
```diff
@@ -4,12 +4,13 @@
 
 return [
     'available_interpreters' => [
+        OpenDialogAi\InterpreterEngine\Interpreters\OpenDialogInterpreter::class,
         OpenDialogAi\InterpreterEngine\Interpreters\CallbackInterpreter::class,
```
stu-greenshoots, line 8: No need to remove this now? Given that the class doesn't work anymore?

stu-greenshoots (commented):
Thanks Elliot - some well thought out work. I have left a number of comments that pose questions. Things I'm thinking in my mind:
- do the before and after callbacks complicate things - could the intent just always be passed in with the interpret method? Is there a use case for the callbacks to do something other than just make the intent available
- With the no-caching option, is there anything to think around using the current intent as part of the cache to always handle this for us?

src/InterpreterEngine/config/opendialog-interpreterengine.php
stu-greenshoots: Agreed that this remains deprecated to leave any existing old config that references it working and review deprecated classes in future releases

src/InterpreterEngine/InterpreterEngineServiceProvider.php
stu-greenshoots: As below,leaving existing config in a working, but deprecated state

src/InterpreterEngine/Interpreters/CallbackInterpreter.php
stu-greenshoots: See comments below

src/InterpreterEngine/Similarity/SimilarityServiceInterface.php
stu-greenshoots: > Thanks Elliot - some well thought out work. I have left a number of comments that pose questions. Things I'm thinking in my mind:
> 
> * do the before and after callbacks complicate things - could the intent just always be passed in with the interpret method? Is there a use case for the callbacks to do something other than just make the intent available
> * With the no-caching option, is there anything to think around using the current intent as part of the cache to always handle this for us?

Having chatted this through with Elliot agreeing to leave caching prevention up to the individual interpreter as narrowing the cache key would ultimately remove the benefit of caching.

Also agree with the before / after callback as opposed to changing method signatures which is much more of a breajing change

stu-greenshoots:
> Thanks Elliot - some well thought out work. I have left a number of comments that pose questions. Things I'm thinking in my mind:
> 
> * do the before and after callbacks complicate things - could the intent just always be passed in with the interpret method? Is there a use case for the callbacks to do something other than just make the intent available
> * With the no-caching option, is there anything to think around using the current intent as part of the cache to always handle this for us?

Having chatted this through with Elliot agreeing to leave caching prevention up to the individual interpreter as narrowing the cache key would ultimately remove the benefit of caching.

Also agree with the before / after callback as opposed to changing method signatures which is much more of a breaking change

src/InterpreterEngine/Service/InterpreterService.php
stu-greenshoots: Elliot will investigate refactoring the interpreter service to not be a singleton

src/InterpreterEngine/BaseInterpreter.php
elliotmassen: The null check in the interpreter service is because these properties are initialised as `null` in the constructor.

https://github.com/opendialogai/core/blob/1216944f408e098866fb01a17c62ceaf86d334f9/src/InterpreterEngine/BaseInterpreter.php#L33-L39

stu-greenshoots (approved):

PR #546, opened 2021-08-02
Handle entities with array of values

istos:
The PR deals with the situation where Dialogflow decides to provide entity results as an array of values rather than a single value. It seems that this happens when Dialogflow detects that the same entity may appear multiple times in a phrase. 

Currently, this short-term solution ignores any values beyond the first one but there is scope to consider a broader solution here that would attempt to cater for multiple values being stored (potentially by using an array attribute in OpenDialog).


src/InterpreterEngine/Interpreters/AbstractNLUInterpreter/AbstractNLUInterpreter.php
```diff
@@ -86,18 +86,40 @@ protected function resolveEntity(AbstractNLUEntity $entity): Attribute
...
+    {
+        $value = '';
+
+        try {
+            if (is_array($resolutionValues)) {
+                $value = $resolutionValues[0];
+            } else {
+                $value =  $resolutionValues;
+            }
+        } catch (\Exception $e) {
```
stu-greenshoots, line 107: I'm not sure this exception could actually be thrown - perhaps if the resolution values were an array but didn't have the first value. I will update this to use Arr::first so we always catch that and remove the catch block

stu-greenshoots (approved):

PR #549, opened 2021-08-06
Adding is_virtual attribute

elliotmassen:
This PR adds an `is_virtual` boolean attribute which indicates whether the previous incoming intent was a virtual intent.


stu-greenshoots (approved):
This is working as expected, thanks 👍

PR #552, opened 2021-08-13
Token based auth

patshone-gsl:
This PR provides an implementation of token based authentication for webchat endpoints using Laravel Sanctum.

The aim of this work is less about actually authenticating requests to the webchat endpoint and more about using the token mechanism to transmit metadata to the OpenDialog backend. 

The primary use-case here is to provide the basis of determining, whether a request to the webchat endpoint should be counted as a production request or otherwise.

The mechanism through which this metadata is submitted to the OpenDialog backend is through the use of Laravel Events.

Events for recording a request for a scenario are dispatched using the ScenarioRequestLoggerMiddleware class.

In order to support the ability to generate tokens automatically a number of events have been created and dispatchers added to provide the main OpenDialog application the ability to respond to Create, Update and Delete events for ConversationObjects.


config/opendialog.php
```diff
@@ -12,4 +12,7 @@
     'migration_publish_dir' => env('MULTI_TENANT', false) ? 'migrations/tenant' : 'migrations',
 
     'virtual_intent_depth' => env('VIRTUAL_INTENT_DEPTH', 5),
+
+    'authenticated_webchat' => env('AUTHENTICATED_WEBCHAT', false),
+    'bot_user' => env('BOT_USER', 'opendialog'),
```
elliotmassen, line 17: What is the `bot_user` config used for?

src/Http/Middleware/ValidateScenarioAccessToken.php
```diff
@@ -0,0 +1,37 @@
...
+    public function handle(Request $request, Closure $next)
+    {
+        $selectedScenario = $request->input('content.user.custom.selected_scenario');
+        if (!$request->user()->tokenCan('scenario:' . $selectedScenario)) {
+            abort(403,
+                'Access denied: the requested scenario -> ' . $selectedScenario .
+                ' does not match the one provided in the token supplied'
+            );
+        }
+        if (!$request->user()->tokenCan('active')) {
```
elliotmassen, line 29: It might be worth adding some constants to store these abilities - less room for error than string values.

elliotmassen (approved):
This looks great - just a couple of minor comments.

PR #557, opened 2021-09-02
Platform configuration fix

elliotmassen:
This PR fixes a bug with platform configurations in which services were loaded in a controller's constructor, which has unexpected affects. This PR resolves those services for `incoming/webchat` endpoint outside of the constructor.


patshone-gsl (approved):
Looks good

PR #562, opened 2021-09-24
Adding data client methods to query incoming transitions

elliotmassen:
This PR adds a `TransitionDataClient` with methods that allow for querying of incoming transitions for conversations, scenes and turns.


stu-greenshoots (approved):
Thanks Elliot, the updates are good and has good test coverage

PR #559, opened 2021-09-09
Adding request ID to /incoming/webchat response meta data

elliotmassen:
This PR adds to the `RequestLoggerMiddleware` to add the current request ID to the response. This PR updates the structure of the `/incoming/webchat` endpoint response to be:

```
{
    "messages": [{...}],
    "meta": {
        "request_id": "od-abcxyz.123"
    }
}
```

PR #567, opened 2021-10-06
Events for conversation logs

stu-greenshoots:
This PR enables the conversation visualiser work within OpenDialog by doing the following:

+ Adds a dependency on `spatie/laravel-event-sourcing` to allow us to easily save events to the database by having them extend `ShouldBeStored`
  + NB config for this placed in OpenDialog app currently. This needs to be reviewed
+ Adds a new DB table 'stored_events' to store each thrown conversation event along with request_id, user_id, timestamp and the event's message to display

+ Adds a new model class for StoredEvents that simplifies the getting of properties and allows the event to return a transition if there is one

+ Adds a bunch of conversation events in the `OpenDialogAi\Core\Conversation\Events` namespace and throws them at various points throughout the conversation engine run

+ Each event will always have the following properties:
  + message - a string representation of the event used to display information to the user
  + status - info, success, error or warning. Used to inform the user if something went wrong
  + request_id - part of the event metadata. Added using the `request_id` bound to the service layer

+ Each event will optionally have the following properties (NB - not all events are thrown against a conversational object):
   + Subject - action, interpretation, intent selection, conditions, messages, attributes, behaviour, transition. If the event is about a particular part of the engine run
   + objectType - scenario, conversation, scene, turn, intent, message. The type of conversational object the event was thrown against
   + objectId - the ID of the conversational object
   + objectName - the name of the conversational object for display

+ Each event stores any additional properties needed to generate the toString, but tries to avoid serialising any whole objects for efficiency

+ Adds a new `describe()` method to attributes that is different from the toString in that it will return the name of the attribute and the value
   + removes the generic toString from attributes where they were repeated

Note - the list total list of events has been reduced from the first iteration to no longer include any selection type events. These ended up not being used in the visualiser and made for a lot of extra noise.

The complete list of events and how they are used can be found in this airtable:
https://airtable.com/appNqBKbyikBfYLC4/tblAIsQ7nMjimhvXP/viwQTFnrD9UiHrwQZ?blocks=hide


stu-greenshoots:
Please note - this PR makes use of https://github.com/opendialogai/core/pull/559 which returns the request id with webchat responses.

PR #574, opened 2021-11-02
Feature/opndlg 1114 template performance

stu-greenshoots:
This PR updates the addFullScenarioGraph method in the scenario data client to minimise the number of queries and mutations made to improve the speed of imports and template generation.

There is a single main mutation that takes the entire structure of a scenario down to its message templates, and this is followed by a subsequent mutation request with 0 - many mutations inside that link intents to their turns. The reduces the number of requests we make, but will still be run as separate mutations inside DGraph.


src/Conversation/DataClients/ScenarioDataClient.php
```diff
@@ -24,13 +34,171 @@ public function addFullScenarioGraph(Scenario $scenario): Scenario
...
+                                MessageTemplate::NAME,
+                                MessageTemplate::MESSAGE_MARKUP,
+                                MessageTemplate::DESCRIPTION,
+                                MessageTemplate::CREATED_AT,
+                                MessageTemplate::UPDATED_AT,
+                                MessageTemplate::CONDITIONS => [
+                                    Condition::OPERATION,
+                                    Condition::OPERATION_ATTRIBUTES,
+                                    Condition::PARAMETERS
+                                ],
```
elliotmassen, line 164: Very minor, but you should be able to replace these with the following (to avoid any potential divergence in future).

```suggestion
                                MessageTemplate::CONDITIONS => Condition::FIELDS,
```

src/Conversation/DataClients/ScenarioDataClient.php
```diff
@@ -24,13 +34,171 @@ public function addFullScenarioGraph(Scenario $scenario): Scenario
...
+                    Scene::TURNS => [
+                        Turn::OD_ID,
+                        Turn::NAME,
+                        Turn::DESCRIPTION,
+                        Turn::INTERPRETER,
+                        Turn::CONDITIONS => Condition::FIELDS,
+                        Turn::BEHAVIORS => Behavior::FIELDS,
+                        Turn::CREATED_AT,
+                        Turn::UPDATED_AT,
+                        Turn::RESPONSE_INTENTS => [
```
elliotmassen, line 139: These intent fields are missing transitions.

elliotmassen (changes requested):

src/Conversation/DataClients/ScenarioDataClient.php
```diff
@@ -24,13 +34,171 @@ public function addFullScenarioGraph(Scenario $scenario): Scenario
...
+                              id
+                              od_id
+                              request_intents {
+                                 id
+                                 od_id
+                              }
+                              response_intents {
+                                 id
+                                 od_id
+                              }
```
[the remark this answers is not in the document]
stu-greenshoots: These have been added so that they fields are returned as an output of the mutation

src/Conversation/DataClients/ScenarioDataClient.php
stu-greenshoots: Thanks, updated

src/Conversation/DataClients/ScenarioDataClient.php
stu-greenshoots: I've added in for consistency, but aware that transitions won't be added from the first mutation result here - and will be added on after the object patching

elliotmassen (approved):

PR #585, opened 2021-12-28
hook conversation crud

goranambar:
This PR creates hooks for Conversation Objects CRUD operations.


stu-greenshoots (approved):
Thanks Goran, the PR looks good. I've have made 2 small changes:
- added in the events for message templates
- removed the updating event for turns on the back of an intent being deleted

PR #587, opened 2022-01-06
Added implementation for supporting file uploads in Open Dialog application

aniekanoffiong:
This Pull Request provides functionality for File Uploads to the Open Dialog application. It also includes config variables that can be used to limit the file upload size and throttling of request to the API endpoint


stu-greenshoots (changes requested):
Hi Aniekan, thanks for this. A few things that I think would improve it:

- Can we have a new service provider for file uploads that registers the service?
- As well as that, can the config file be moved here alongside it?
- Files should be renamed on saving, it would be good to centralise this logic here. For renaming, let's use the current timestamp and the original file extension 
- Can the registration of s3 or local file upload clients be based on a config value instead of multi-tenant

src/Traits/FileRename.php
```diff
@@ -0,0 +1,24 @@
...
+
+trait FileRename
+{
+
+    public function renameFile(UploadedFile $file)
+    {
+        // Adding random string to prevent same timestamp files collision
+        $randomString = Str::random(5);
+
+        $currentTimestamp = Carbon::now()->format('Y-m-d\TH:i:s.uP');;
```
stu-greenshoots: @patshone-gsl - I'm wondering about this name format - files come out with a name along the lines of:
5ZeLa-2022-01-12T11:55:34.216771+00:00.jpg

I think a nicer format would be without the separators and timezone info. Any thoughts?

patshone-gsl (approved):

PR #588, opened 2022-01-07
Allowing event persisting condition to be overridden

elliotmassen:


src/Conversation/Events/Storage/StoredEventRepository.php
```diff
@@ -31,7 +31,7 @@ public function persist(ShouldBeStored $event, string $uuid = null, int $aggrega
      * @param $event
      * @return bool
      */
-    private function shouldPersist($event): bool
+    public function shouldPersist($event): bool
```
stu-greenshoots, line 34: Would protected work here?

PR #589, opened 2022-01-13
Moved WebchatSettings Model and migrations from webchat package to core

aniekanoffiong:
This Pull Request 

- moves the Model for Webchat Setting from the webchat package to the core package 
- moves the migrations for creating the DB tables. 
- adds the migration for renaming the webchat_settings table to component_settings and adding a new field (component_id) for storing the component_id for the specific settings
- renames the WebchatSetting Model to ComponentSetting


src/Console/Commands/ComponentSettings.php
```diff
@@ -0,0 +1,163 @@
...
+                        [
+                            'display_name' => $details[ComponentSetting::DISPLAY_NAME] ?? $settingName,
+                            'description' => $details[ComponentSetting::DESCRIPTION] ?? null,
+                            'display' => $details[ComponentSetting::DISPLAY] ?? true,
+                            'section' => $details[ComponentSetting::SECTION] ?? null,
+                            'subsection' => $details[ComponentSetting::SUBSECTION] ?? null,
+                            'type' => $details[ComponentSetting::TYPE] ?? ComponentSetting::STRING,
+                            'parent_id' => $parentId,
+                            'sibling' => $sibling,
+                            'component_id' => $details[ComponentSetting::COMPONENT] ?? ComponentSetting::WEBCHAT,
```
elliotmassen, line 58: Does this mean that each array element in config has to specify the component ID? If so, could we somehow key each addition to `opendialog.component_settings` with the component ID instead?

elliotmassen (changes requested):
Looking good, just a few comments:

 - Let's squash these migrations in a single migration
 - The constants in `ComponentSetting` from `GENERAL` to `NUMBER_OF_MESSAGES` and `BOT` to `END_CHAT_CONFIRMATION_NEGATIVE` should remain in the webchat package. The two platform constants should remain in their respective packages.

PR #591, opened 2022-01-13
Feature/opndlg 1333 long text message

stu-greenshoots:
This PR makes the required changes in core for bringing the LongTextMessage back into the fold.

There are 3 main updates here:

- Long text message updated to allow for `text` and `attribute_name` (updates in the webchat message formatter and associated classes)
- Creation of a new message type `long_text_edit` that handles saving edit clicks to the messages table without affecting conversation. The Webchat Sensor will identify the utterance, persist to the message table, and then throw the correct Exception to ensure it is not interpreted
- Character Limit enforced as an int rather than string.

Tests have been updated for all of the above.

The handling attribute name has been placed in the main WebchatMessage so that all other message types can also make use of it


patshone-gsl (approved):
Validated locally

PR #593, opened 2022-01-19
Show hide features

patshone-gsl:
This PR refactors how and when PlatformComponents are provided to the Reflection API. It does so by moving the filtering of excluded platforms from the register method of the `PlatformEngineServiceProvider` to the `all`, `has` and `get` methods of the `PlatformComponentService`

This change has the impact such that when requests are made for PlatformComponents are made, they are filtered after middleware has run rather than before, meaning that we can if needs be, adjust which Platforms are made available to the application.


patshone-gsl:
Closing this PR in preference to the approach here - https://github.com/opendialogai/core/pull/596

PR #599, opened 2022-01-27
Collection attribute support

elliotmassen:
This PR adds support for collection attributes. Although more complex attribute support is still ongoing, since `1.0` OpenDialog has set out three kinds of attributes:

 - **Scalar**: These are the most common and well supported kinds of attributes, like string, int, boolean, timestamp, etc.  A scalar attribute maps an attribute ID to a string attribute value. For example `my_counter: 5`.
 - **Composite**: These only have partial, programmatic support currently, and are only used internally for utterances. A composite attribute maps an attribute ID to mapping of attributes. For example `address: {building_no: 1, street: "Test Street", city: "London", postcode: "ABC XYZ"}`.
 - **Collections**: This PR adds support for this kind of attribute. A collection attribute maps an attribute ID to an array of attribute _values_.  For example `shopping_list: ["apples", "bananas", "carrots"]`. All standard scalar attributes have a corresponding collection attribute, like string collection, int collection, etc.

As emphasised above, the introduction of collection attribute support places further importance on the distinction between an attribute and an attribute value. In a collection attribute, we have just one actual attribute, but zero-or-many attribute values. This is what marks the difference between a collection and a composite.

This PR adds support for collection attributes in message filling, operations, actions, contexts & interpreters. Specific values from a collection can be indexed via a zero-based array accessor notation: `my_collection[0]`. Collections can also be accessed as a whole like scalar attributes: `my_collection`.

As OpenDialog's handling of types currently places more emphasis on attributes, rather than attribute values (eg. an action requires inputs & outputs identified by attribute ID, and therefore attribute type, rather than attribute value type), this PR also adds the ability to index a specific value in a collection attribute and treat it is a attribute via the `getValueWrappedInAttribute` method. In this case the wrapped attribute ID will include the accessor notation.

There was some pre-existing, incomplete functionality regarding accessors which was centred around the composite use-case, ie `my_composite[my_sub_attribute_id]`. As mentioned above, composite support is partial and limited to programatic use, therefore the scope of this PR has been to only ensure that accessor notation works for collections, as full composite support is still awaiting completion.


src/AttributeEngine/AttributeBag/HasAttributesTrait.php
```diff
@@ -23,7 +24,25 @@ trait HasAttributesTrait
...
+
+        if (count($accessors) < 1) {
+            // The attribute is present, and no accessors were specified
+            return true;
+        }
+
+        // The attribute is present, and accessors were specified
+
+        /** @var CollectionAttribute $collectionAttribute */
+        $collectionAttribute = $this->getAttribute($id);
```
stu-greenshoots, line 43: What happens here if an accessor is supplied by mistake and the underlying attribute is not actually a collection attribute? Does it cause an error if `getValueAt` doesn't exist?

src/AttributeEngine/AttributeBag/HasAttributesTrait.php
```diff
@@ -75,11 +115,11 @@ public function setAttributes($attributes)
      */
     public function setAttribute(string $attributeName, $value): Attribute
     {
-        // If the attribute exists update its value
         if ($this->hasAttribute($attributeName)) {
             $attribute = $this->getAttribute($attributeName);
```
stu-greenshoots, line 119: Is there anyway the attribute name could include accessors here? If

src/AttributeEngine/Attributes/AbstractCollectionAttribute.php
```diff
@@ -53,4 +71,76 @@ public function getValue()
...
+        return new Map($this->values);
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function removeValue(int $index): void
+    {
+        unset($this->values[$index]);
+        $this->values = array_values($this->values);
```
stu-greenshoots, line 94: Is this done to reset the array?

src/AttributeEngine/AttributeBag/HasAttributesTrait.php
```diff
@@ -46,7 +65,28 @@ public function hasAllAttributes($attributes): bool
...
     {
-        $this->attributes->put($attribute->getId(), $attribute);
+        $id = $attribute->getId();
+        $accessors = Util::parseStringToAccessors($id);
+        $id = Util::removeAccessors($id);
+
+        if (count($accessors) > 0) {
+            if (!$this->hasAttribute($id)) {
+                // If the parent collection doesn't exist, create it
+                $collectionClass = $attribute::getCollectionClassType();
```
stu-greenshoots, line 75: This collection class could be null for any non-string type - will this be handled?

src/AttributeEngine/Util.php
```diff
@@ -8,51 +8,54 @@
  */
 class Util
```
stu-greenshoots, line 9: I feel like we can name this better :)
AttributeNameUtils, AttributeUtils ..?

stu-greenshoots (changes requested):
Awesome work Elliot, I've had a little play and things seem to be working nicely. I've added some comments as I looked through the code, but can't find anything major in there. Happy to have a chat through

src/AttributeEngine/Attributes/AbstractCollectionAttribute.php
elliotmassen: Yes that's right - the assumption is that a collection attribute should maintain sequential indexes. Therefore removing an index causes all subsequent indexes to shift down to accommodate.

src/AttributeEngine/AttributeBag/HasAttributesTrait.php
elliotmassen: Good catch, I've made a change so that in that case the attribute will not be added.

src/AttributeEngine/AttributeBag/HasAttributesTrait.php
elliotmassen: Yes - in fact it's used in `testExplicitlyRemoveCollectionItem` via the `removeAttribute` method. In this case the `hasAttribute`, `getAttribute` and `addAttribute` used all account for the accessors and handle them correctly.

src/AttributeEngine/AttributeBag/HasAttributesTrait.php
elliotmassen: I've just made a change which will cause it to ignore the accessors if the attribute isn't a collection attribute. This means that it will treat `my_int_attribute[0]` like `my_int_attribute`. It feels like this will require some further thought when we add full support for composites.

stu-greenshoots (approved):

PR #600, opened 2022-02-01
Check for userId in ScenarioRequestLoggingMiddleware

patshone-gsl:
This PR adds an additional check to the ScenarioRequest logging middleware so that we only create an event if we have a valid userId


src/Http/Middleware/Tests/ScenarioRequestLoggerMiddlewareTest.php
```diff
@@ -0,0 +1,105 @@
...
+use Laravel\Sanctum\HasApiTokens;
+use Laravel\Sanctum\PersonalAccessToken;
+use OpenDialogAi\Core\Http\Events\ScenarioAccessRequestEvent;
+use OpenDialogAi\Core\Http\Middleware\ScenarioRequestLoggerMiddleware;
+use OpenDialogAi\Core\Tests\TestCase;
+use Illuminate\Foundation\Auth\User;
+
+class ScenarioRequestLoggerMiddlewareTest extends TestCase
+{
+    public function setUp(): void
```
stu-greenshoots, line 19: not a big thing, but not needed here I think?

stu-greenshoots (approved):
Just 1 minor comment in the test, but all good otherwise 👍

PR #603, opened 2022-02-08
Added property for handling property hiding for configuration objects

aniekanoffiong:
This Pull Request implements logic for retrieving properties to be hidden in the configuration response for requests to retrieve component configurations


elliotmassen (approved):
This looks good! Just one small suggestion below.

PR #608, opened 2022-02-11
Adding deleteUser method

elliotmassen:
This PR gives the ability to completely remove a user's data from the graph.


patshone-gsl (approved):
Following review with Elliot

PR #613, opened 2022-02-23
OPNDLG-1362 Fixes the perseverance of intents order on duplicate scenario

goranambar:
Fixes the perseverance of intents order when exporting/importing/duplicating scenarios.
The way I tried this is exporting existing scenario and checking that intents have correct order in the exported json. Then I have duplicated the scenario and exported that, and checked that the intents there have the same order, too.


elliotmassen (changes requested):
I've tested this locally with importing a scenario and then duplicating it, and I'm still seeing the intents ordered differently unfortunately.

We will also require these ordering fixes for message templates as their ordering is important to the conversation engine, and equally part of this issue. @stu-greenshoots I'm wondering if it'd be worth expanding the changes from https://github.com/opendialogai/core/pull/574/commits/0a7eb0df5bda16a058734c36e48e9eb0add9dafc to include any conversation object. That way even the ordering of conversations, scenes and turns are exactly the same - as well as intents & message templates.

goranambar:
> I've tested this locally with importing a scenario and then duplicating it, and I'm still seeing the intents ordered differently unfortunately.

Hey @elliotmassen , can you tell me how do you test this? How do you see the intents ordering?

goranambar:
> We will also require these ordering fixes for message templates as their ordering is important to the conversation engine, and equally part of this issue. @stu-greenshoots I'm wondering if it'd be worth expanding the changes from [0a7eb0d](https://github.com/opendialogai/core/commit/0a7eb0df5bda16a058734c36e48e9eb0add9dafc) to include any conversation object. That way even the ordering of conversations, scenes and turns are exactly the same - as well as intents & message templates.

Practically this is what needs to be done. Intents are ordered correctly within the turns, but the conversation, scenes and turns are not.

elliotmassen:
> > I've tested this locally with importing a scenario and then duplicating it, and I'm still seeing the intents ordered differently unfortunately.
> 
> Hey @elliotmassen , can you tell me how do you test this? How do you see the intents ordering?

I'm seeing it when I have an export with a turn with two request intents: A, B (in that order). When I import them, the turn has request intents ordered: B, A. Then I duplicate that scenario and they are now ordered A, B. Etc, etc.

src/Conversation/DataClients/Serializers/Normalizers/ImportExport/MessageTemplateNormalizer.php
```diff
@@ -46,6 +51,10 @@ public function denormalize($data, string $type, string $format = null, array $c
             $messageTemplate->setMessageMarkup($data[self::MESSAGE_MARKUP]);
         }
 
+        if (isset($data[self::ORDER])) {
+            $messageTemplate->setMessageMarkup($data[self::ORDER]);
```
elliotmassen, line 55: Think this is a mistake?

```suggestion
            $messageTemplate->setOrder($data[self::ORDER]);
```

elliotmassen (changes requested):

src/Conversation/DataClients/Serializers/Normalizers/ImportExport/MessageTemplateNormalizer.php
stu-greenshoots: thanks, yes, that is a mistake!

PR #615, opened 2022-03-04
Fixes the createFromFormat error

goranambar:
Fixes the DateTime creation from LARAVEL_START const not properly initialized


stu-greenshoots (changes requested):
Hey Goran, I think this will stop the error occurring, but will end up with the wrong times being logged.

After chatting to Pat, he remembered that we had this issue before with microtime where the precision isn't added if it hits exactly on a second. 

We have fixed it with this approach before:
https://github.com/opendialogai/core/commit/49197fee8f9a4a64a1f2a3d9b9f9e57ec97568cd

Could you take a look and update to something like this please

stu-greenshoots (approved):

PR #616, opened 2022-03-08
Corrects some validation messages

goranambar:
Corrects some validation messages wording.


src/ResponseEngine/Rules/MessageXML.php
```diff
@@ -227,7 +227,7 @@ public function passes($attribute, $value)
                                 return false;
                             }
                             if (empty((string)$element->name)) {
-                                $this->setErrorMessage('Form message elements must have "name"');
+                                $this->setErrorMessage('Attribute "name" is required for form fields');
```
stu-greenshoots, line 230:
```suggestion
                                $this->setErrorMessage('"Attribute name" is required for form fields');
```

stu-greenshoots: I think the double-quote should go around the whole thing so it doesn't look like it is looking for an attribute called "name"

stu-greenshoots (changes requested):
Just a suggested little tweak to the wording

stu-greenshoots (approved):

PR #618, opened 2022-03-10
Handling messages from external NLU

goranambar:
Adds <message> tag to DialogFlow messages. Handles subsequent messages after attribute messages.


stu-greenshoots (approved):
I have checked this out locally on the Perfect Fit scenario and confirmed that it works as expected.

Where the attribute was previously used directly in the message (but not wrapped in an OpenDialog message type) `{user.dialogflow_message}`, it now must be used in an attribute message like this:

```
<attribute-message>
  user.dialogflow_message
</attribute-message>
```

The context and attribute name should not be wrapped in curly braces so that it is not replaced by attribute filling.

Merging of this branch will need to happen in conjunction with updating any affected messages

PR #619, opened 2022-03-17
DGraph Query Caching

stu-greenshoots:
This PR adds a layer of caching to the DGraph client to reduce the overall number of HTTP calls made to DGraph during requests.

The update adds a new CacheHelper class that does the following:

+ generates a unique but reproducible cache key by hashing the full query string along with any variables. This key will always be the same for the same query using the same variables.

+ Check if a query should be cached. Mutations and any query relating to users, attributes or contexts are not valid for caching as the results are expected to change frequently

+ Check if there is something in cache for a cache key

+ Get values from the cache for a cache key

+ Add items to the cache given a cache key and the data to persist. A cache TTL value is taken from the config for defining how long values should live in the cache

+ Bust caches given a query. Checks if the query is from the list of non cachable mutations (anything to do with users, contexts or attributes) and avoids busting cache if so. Otherwise if the query is any other mutation (ie updating data) the entire query cache is busted to ensure that updated data is returned on the next call

Caching is achieved using [Laravel tagged caching](https://laravel.com/docs/9.x/cache#cache-tags) meaning that only certain cache drivers are supported. This needs to be reflected in the apps `CACHE_DRIVER` env.

This allows all query cached to be tagged meaning it can all be flushed in one go when required.

The DGraph client `request` method flow has been updated to now do the following:

+ First try to bust the cache. This ensures any data is cleared out if needed. This will happen when there are any updates to conversation objects (Scenarios -> Message Templates) but not for any user-related items (user, attributes and contexts)
+ Check if the query should be cached. If so, generate a cache key, check if there is an item in the cache and return it if so
+ If there is nothing in the cache, the query is made, as usual, then if the query is cachable, the result is added to the cache. The new cache value should then be used the next time the same query is made with the same variables.

3 new environment variables are needed for this workflow:

`DGRAPH_QUERY_CACHE_ON` - A global switch for query caching. If false, queries will never be cached
`DGRAPH_QUERY_CACHE_TAG` - The tag to use when caching (defaults to `scenario_data`)
`DGRAPH_QUERY_CACHE_TTL` - Time to live for items in the cache in seconds. Defaults to 3600 (1 hour)

Running some loads tests to compare with and without caching:

With caching turned off:
image

With caching turned on:
image


patshone-gsl:
I'm interested in knowing how this works in saas as it looks like the saas package also uses tags to separate the tenants cache. Below is from `config\tenancy.php`

```
    /**
     * Cache tenancy config. Used by CacheTenancyBootstrapper.
     *
     * This works for all Cache facade calls, cache() helper
     * calls and direct calls to injected cache stores.
     *
     * Each key in cache will have a tag applied on it. This tag is used to
     * scope the cache both when writing to it and when reading from it.
     *
     * You can clear cache selectively by specifying the tag.
     */
    'cache' => [
        // This tag_base, followed by the tenant_id, will form a tag that will be applied on each cache call.
        'tag_base' => 'tenant',
    ],
```

stu-greenshoots:
@patshone-gsl : I've made the following checks and updates:

- Confirmed that the tenancy provider uses both the tenant cache tag as well as any other tag that we provide using the cache facade tag method
- Confirmed that with `DGRAPH_QUERY_CACHE_ON` env set to false, a cache driver that doesn't support tags such as `file` can be used with no issues - if `DGRAPH_QUERY_CACHE_ON` is set to true, it will though
- I have made an update to the standard development environment to include redis so this that cache driver can easily be used (https://github.com/opendialogai/opendialog-dev-environment/pull/30)
- Created a PR on OpenDialog app that uses this branch of core and updates the example env  (https://github.com/opendialogai/opendialog/pull/463)

patshone-gsl (approved):

PR #620, opened 2022-03-17
Conversation event dispatching conditioning

stu-greenshoots:
This PR updates the BaseConversationEvent to condition the dispatching of the event. Since so many conversation events can be generated through the flow of a conversation, it adds overhead to dispatch events that are never intended to be dealt with.

Events will not be dispatched if they are either globally turned off via the config, if there is no active auth user or if the user's email address matches the bot user registered in config. This should mean that events are dispatched only for active admin users and not for any requests from outside of the admin interface.

This approach is valid for webchat calls where it is known if the call originated from our admin interface or not - but it's not valid for other platforms where we have no way of telling. In this case, events will never be dispatched


stu-greenshoots:
Agreed with @patshone-gsl  to not merge this at this time until we can see a way to improve for Alexa calls as well

PR #621, opened 2022-03-22
Add batch dgraph updates

goranambar:
Adds batch updates to intents.


stu-greenshoots:
unfortunately, this does not have the effect we wanted and the slow query times at DGraph still persist. The change in code makes it harder to read, and so decided not to merge this in.

PR #625, opened 2022-03-30
Prevent attempt to bust cache if caching is not turned on

stu-greenshoots:
This PR updates the CacheHelper to avoid cache-busting if caching is not turned on


goranambar (approved):
Tried it locally and it works OK

PR #632, opened 2022-05-06
Add interfaces to data clients

goranambar:
Gets all of the Data Clients behind interfaces. This is going to make the migration easier, and also it is better for future changes.


stu-greenshoots:
Closing as future PR covers all the work

PR #643, opened 2022-06-28
Move the request/response logger code into its own service

goranambar:
Extract that code into a separate helper class `RequestLoggerHelper` which can be used in `RequestLoggerMiddlewareand` also in webchat.


elliotmassen (approved):
Looks good to me 👍 Happy for this to be merged.

PR #647, opened 2022-07-07
Fixes the intent and messages ordering bug

goranambar:
Adds getNextOrderNumber for ordering fix and reads the complete intent object, with message templates


elliotmassen (approved):
This looks good! The only (very, very minor) thing I'm wondering is if the method should be `getHighestOrderNumber` and that incrementation becomes the responsibility of the caller instead? That way the method would be agnostic to its use, and could be used for other purposes too.

src/Conversation/DataClients/Relational/ConversationDataClient.php
```diff
@@ -694,7 +694,8 @@ public function getIntentByUid(string $intentUid): Intent
         $conversation = $this->createConversationObject($conversationModel, $scenario);
         $scene = $this->createSceneObject($sceneModel, $conversation);
         $turn = $this->createTurnObject($turnModel, $scene);
-        $intent = $this->createIntentObject($intentModel);
+        $intent = $this->createIntentObjectTree($intentModel,
```
elliotmassen, line 697: Is this related to this PR?

goranambar: Kind of yes, because that function is used when binding value of Intent in the routes and if the whole tree isn't read, then the MessageTemplates aren't read and if they are not read then adding a new MessageTemplate will always have its ordering at 0.

PR #649, opened 2022-07-08
Handback support

elliotmassen:
This PR is a continuation of the work from #644, and allows an external conversation system to "handback" responsibility to OpenDialog. This is facilitated through a new conversation engine entrypoint called `runHandback`. To enable this handback, it has become necessary to store the current message template ID in conversational state. This allows the handback functionality to pick up where it left off, and process any remaining messages in the message template, as well as any virtual intents.


goranambar (commented):
The code looks OK. Chatwoot communication is OK. There are some tests failing.

I Still need to wrap my head around what needs to happen after handback support, but I guess we're going to revisit this PR once it is completely finished.

goranambar:
Looks OK now.

goranambar (approved):

PR #651, opened 2022-07-13
Fixes the bug that Response Intent doesn't get created

goranambar:
The problem was detected in `TurnsController::storeTurnIntentAgainstTurn`. Saved intent didn't contain the parent scenario.
Now, the problem arises from `Relational\ConversationDataClient::getTurnByUid` because it is not setting the scenario scene/conversation parents. This is used in the `RouteServiceProvider` for `turn` binding. Setting the conversation/scenario should fix the problem


patshone-gsl (approved):
Validated locally - I can see that this fixes the issue

PR #655, opened 2022-07-25
Adds addAttributeBySubtype method to Composite

goranambar:
Adds addAttributeBySubtype method for composite attributes and updates the AttributeResolver::getAttributeFor


elliotmassen (commented):
I'll revisit this once the composite accessor work is complete - so that we can assess their overlap.

PR #660, opened 2022-08-08
Composite attribute accessor support

elliotmassen:
This PR adds support for accessing composite attributes via `[]` notation. It follows conventions set in #599 by ensuring that each nested sub-attribute's ID is the absolute ID. This has led to changes in some pre-existing, programmatic composite attributes (user, utterance & history record), which have been accounted for by updating the relevant helper methods.

Sub-attributes of composites can be accessed like so:

`user.my_composite['my_nested_sub_attribute']`

The use of single quotes is required, and leaves future scope for dynamically accessing sub-attributes if necessary.


goranambar (commented):
Hey Elliot, nice work! I can see that a lot is happening in here. A lot of fiddly work.

Mainly, I was concerned about the composite and inferred composite attributes working and I can see that the tests are updated and they are working OK with accessors support.

Will look through this again once it is out of draft status. And all of the tests are corrected, off course :)

PR #661, opened 2022-08-17
Sub-attribute message filling fix

elliotmassen:
This PR ensures that sub-attribute accessors with encoded quotes work during message filling. Before this PR when messages contained an attribute replacement statement like `{user.my_attr['sub_attr']}` and encoded as `{user.my_attr[&#x27;sub_attr&#x27;]}`, the message matcher was not able to correctly find the attribute. This PR replaces the single quote encodings to allow the sub-attribute to be found.


goranambar (approved):
Looks OK Elliot.

PR #664, opened 2022-09-06
OD Translate Spike

stu-greenshoots:
This branch contains spike work and is not to be merged.

This spike aims to experiment with translation strings in message templates.

The main logic happens in a new class `MessageTemplateTranslator` that can replace translatable fields in message XML to be used when saving a message template and methods to exchange the translation strings for their translations for display.

It uses Laravel's Blade template engine for rendering an entire message string and performing all translations at once - this relies on the translation keys being wrapped in Laravel's blade directive for translation `@lang()`


src/ResponseEngine/Formatters/Webchat/WebChatMessageFormatter.php
```diff
@@ -88,7 +89,7 @@ public function getMessages(string $markup): OpenDialogMessages
     {
         $messages = [];
         try {
-            $message = new SimpleXMLElement($markup);
+            $message = new SimpleXMLElement(MessageTemplateTranslator::translateMessageMarkup($markup));
```
stu-greenshoots, line 92: We would probably want this to happen a level above the interface's formatter so it works for all interfaces

PR #667, opened 2022-10-06
Hand-to-system handback fix

elliotmassen:
This PR fixes a bug with the handback method by updating conversation the method to retrieve the user attribute directly.


goranambar (approved):
I've tested it locally and I can see that after agent selects `Close Conversation` I'm being handed back to system.

PR #668, opened 2022-10-07
Null value attribute handling

stu-greenshoots:
This PR updates the HasAttributesTrait to return true for `hasAttribute` even if the value is null.

The issue arose in a project where an action returns null values for a set of attributes expecting that value to be saved to the user context, resetting current values. A recent change to the trait meant that `hasAttribute` started returning false if the attribute was in the bag, but had a null value. This manifested itself in a exception being thrown by the `getAttribute` method.

This change reverts that bit of logic and adds a test to highlight.

It looked like there could have been a related issue in the ActionPerformer if the values in `$contextAttributeMap` differed from `$resultAttributes` in the `saveAttributesToContexts`, however after some more testing, it seems this is not possible


elliotmassen (approved):
Look good and works as expected. It also could be worth adding a new that asserts that a bag with a null'd string attribute returns true for `hasAttribute`.

PR #669, opened 2022-10-13
Spike/opndlg 1636 scenario id alias

Lintume:
We want to add aliasing to scenario ID. What I found:

I found all the places where scenario id using from request when webchat loads with embed code.

**In webchat first time:**

`\OpenDialogAi\Webchat\Http\Controllers\WebchatSettings::getScenarioIdFromRequest`

**in core, a chain of places of using scenario id from request:**

`src/Http/Middleware/ValidateScenarioAccessToken.php:28
src/Http/Middleware/ScenarioRequestLoggerMiddleware.php:25`
`src/Http/Controllers/WebchatIncomingController.php:42`
`src/SensorEngine/Sensors/WebchatSensor.php:40`

Because of many places of taking scenario id from request, I've decided to create middleware in the core that transforms Scenario ID inside of the request and connect this middleware to 2 routes in webchat project (webchat-config route) and in the core project (incoming/webchat route)
Because in these 2 endpoints the scenario id is located in different paths of the request - in the middleware written an awkward 'if' for $request->path(). maybe we need to divide this middleware into 2, or smth like this.

*How to use*
- `composer install`
- found in core src/Http/Middleware/TransformScenarioId.php and set up scenario_id. instead of '210'. 
- Change embed code, set up `selected_scenario` to `custom_scenario`.
- clean the cache and reload the page with embed code.

*Related PRs:*
[webchat](https://github.com/opendialogai/webchat/pull/394)
[core](https://github.com/opendialogai/core/pull/669)
[OD-saas](https://github.com/opendialogai/opendialog-saas/pull/159)

*Important note*
If we change the scenario ID in the request - in any logs will put "true" scenario id (not alias) can bring some problems.


Lintume:
We want to add aliasing to scenario ID. What I found:

I found all the places where scenario id using from request when webchat loads with embed code.

**in webchat first time**

\OpenDialogAi\Webchat\Http\Controllers\WebchatSettings::getScenarioIdFromRequest

**in core, a chain of places of using scenario id from request:**

src/Http/Middleware/ValidateScenarioAccessToken.php:28
src/Http/Middleware/ScenarioRequestLoggerMiddleware.php:25
src/Http/Controllers/WebchatIncomingController.php:42
src/SensorEngine/Sensors/WebchatSensor.php:40

Because of many places of taking scenario id from request, I've decided to create middleware in the core that transforms Scenario ID inside of the request and connect this middleware to 2 routes in webchat project (webchat-config route) and in the core project (incoming/webchat route)
Because in these 2 endpoints the scenario id is located in different paths of the request - in the middleware written an awkward 'if' for $request->path(). maybe we need to divide this middleware into 2, or smth like this.

*How to use*
- found in core src/Http/Middleware/TransformScenarioId.php and set up scenario_id. instead of '210'. 
- Change embed code, set up `selected_scenario` to `custom_scenario`.
- clean the cache and reload the page with embed code.

Related PRs:
[webchat](https://github.com/opendialogai/webchat/pull/394)
[core](https://github.com/opendialogai/core/pull/669)
[OD-saas](https://github.com/opendialogai/opendialog-saas/pull/159)

stu-greenshoots (changes requested):
Hey @Lintume, thanks for this spike work. You've done a great job finding the places used, and I think the approach of updating the request as this is what is used through the flow works well.

I wonder if rather than a middleware that tries to handle any requests, the update could be made to request object used by the controller using something like prepare for validation:
https://laravel.com/docs/9.x/validation#preparing-input-for-validation. This would mean needing to create a request class for webchat config, but I think would be a bit cleaner.

I would like to see how we could separate out the code that changes the scenario ID to the aliased one as well. I think this would work well as a Facade in core/conversation engine and usable as something like `ScenarioAlias::getReplacement($scenarioId). This way the code is reusable in the voice package and any other places that we want to alias the scenario ID

Lintume:
Hey @stu-greenshoots,
I've added aliasing for appKey too.

_How to use_

- php artisan migrate from od-saas
- composer install
- create new alias: found in postman 'Opendialog Admin' new folder 'Scenario Id Aliasing', and run 'Create new Alias' endpoint with scenario_id_alias and scenario_id parameters. you'll get new appKey in response ('access_token_alias').
- Change embed code, set up selected_scenario to the your new scenario_id_alias and app key to access_token_alias that you got from the previous step.
- clean the cache and reload the page with embed code.

_Note_:
you can also update, delete, and get all aliases by using postman endpoints in the same folder.

database/migrations/2021_09_10_095958_create_stored_events_table.php
```diff
@@ -22,4 +22,14 @@ public function up()
             $table->unique(['aggregate_uuid', 'aggregate_version']);
         });
     }
+
```
stu-greenshoots, line 25: hmm, was this change necessary, or something to do with the migrations in saas?

src/Conversation/ScenarioAlias.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+
+namespace OpenDialogAi\Core\Conversation;
+
+use OpenDialogAi\Core\ScenarioIdAlias;
+
+class ScenarioAlias
+{
+    public function getOriginalScenarioId(string $scenarioId): string
```
stu-greenshoots, line 10: When we put this into code to be merged, let's add some phpdoc here to explain what the methods are going to do and return

stu-greenshoots, line 10: Could the parameter names be a bit clearer? For this method and the getOriginalToken method, it is the alias we are passing in. Should the name of the parameter be `$alias`? or `$scenarioAlias`?

src/Conversation/Facades/ScenarioAlias.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace OpenDialogAi\Core\Conversation\Facades;
+
+use Illuminate\Support\Facades\Facade;
+
+class ScenarioAlias extends Facade
```
stu-greenshoots, line 7: See our other facades where we add `@method` helpers to make things easier

routes/web.php
```diff
@@ -12,7 +13,7 @@
     \Illuminate\Routing\Middleware\SubstituteBindings::class,
 ]], function () {
     if (config('opendialog.core.authenticated_webchat')) {
-        Route::middleware('auth:sanctum')->post(
+        Route::middleware([TransformScenarioId::class, 'auth:sanctum'])->post(
```
stu-greenshoots, line 16: I can see this was a problem from before, but having 2 middleware declarations on this method is a bit confusing. Could you have it so it's only 1? Either all in the `::middleware` at the start or the `->middleware` after

routes/web.php
```diff
@@ -12,7 +13,7 @@
...
             '/incoming/webchat',
             'OpenDialogAi\Core\Http\Controllers\WebchatIncomingController@receive'
         )->middleware([
```
stu-greenshoots, line 19: I think the route defined in the `else` block will need the transform middleware also?

stu-greenshoots (changes requested):
This looks good to me and I like that the middleware is simplified to only deal with 1 method.

I added some comments on the code style, up for discussion :)

database/migrations/2021_09_10_095958_create_stored_events_table.php
Lintume: I just found this migration without a down function, which I thought was wrong, it caused problem when you rollback migrations

src/Conversation/ScenarioAlias.php
Lintume: it can be not only an alias... here can be passed the original scenario id too, in this case it will be returned without changes

Lintume: added

src/Conversation/Facades/ScenarioAlias.php
Lintume: added

routes/web.php
Lintume: fixed

routes/web.php
Lintume: added

PR #671, opened 2022-10-26
Fix for denormalising composite attributes nested within a collection attribute

elliotmassen:
This PR ensures that composite attributes and (their sub attributes) that are nested within collection attributes retain their attribute ID's when denormalised. Previously the attribute denormaliser presumed that the children of a collection attribute did not need to have a specific attribute ID. Since the introduction of composite attributes in [1.20](https://github.com/opendialogai/core/releases/tag/1.20.0), composites and their children require an absolute ID.


goranambar (approved):
Nice work Elliot! Really sweet :+1: 

I've tested it locally by disabling the overriding of `UserContext` in `Peppercorn` code-base and I can see that the complex `Quote` object gets serialized and de-serialized properly.

PR #673, opened 2022-11-08
Adding action callback usage to action engine reflection

elliotmassen:
This PR updates the reflection of the action engine to include whether each action component uses before/after perform callbacks.

Currently action callbacks are defined in an action's constructor, using instance properties. This means that it is not possible know whether an action uses callbacks without instantiating the action. This posed a problem for the reflection of action components, as we rely on the ability to reflect the component without needing to instantiate it. Therefore it became necessary to introduce a set of parallel static properties/methods to indicate whether an action uses the callbacks. It is entirely possible that an action developer may not use these new static properties (as they aren't required for the actual use of the callbacks), however this will lead to an inaccurate reflection of the action. To prompt their use, a warning will be logged if callbacks are set without the relevant static property indicating it.


goranambar (approved):
Tested it in `OD` and it works as expected.

PR #675, opened 2022-11-14
ScenarioExists rule at OD-SAAS not working properly

Lintume:
Resolving note: If you do a comparison in MySQL between an integer and a string, the string is converted to an integer. 
I've created a trait `StrictIntegerColumns` that analyses the builder object before performing a query. it analyses all wheres on the integer type values, and throws an error if it's not. Properties for the model that should be int type are listed in property `$strict_integer_columns`. If `$strict_integer_columns` is not defined in the model but the trait is defined - working for the `id` column.
I added this trait to ConversationObjectBase and User models.
This solution also treats problems with laravel routing. (when ids are passed as part of the URL)
This trait should be added mandatory to all models if we want to avoid the ability to make queries with described in task troubles.
[Task in Jira](https://opendialog.atlassian.net/jira/software/projects/OPNDLG/boards/13?assignee=6318527c09bc6014ea8c8e37&selectedIssue=OPNDLG-1727)

Related PR:
[OD-Saas](https://github.com/opendialogai/opendialog-saas/pull/170)


src/Traits/StrictIntegerColumns.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        $wheres = $builder->getQuery()->wheres;
+        foreach ($wheres as $where) {
+            if (!isset($where['column'], $where['value'])) {
+                continue;
+            }
+            foreach ($columns as $column) {
+                if ($where['column'] == $column && !is_numeric($where['value'])) {
+                    //if the string is not an integer, by definition it cannot be found in a column with int values
+                    throw new ModelNotFoundException(sprintf('Entity with %s %s not found',
```
patshone-gsl, line 56: There's something here wrt the exception thrown (ModelNotFoundException), that I'm not sure is quite right. If we're throwing an exception here we'd want to be catching it and handling elsewhere and in the first instance with the data clients. In understand why you'd want it to be more generic but I think in this case maybe we want to continue to throw a ConversationNotFoundException and maybe change the name of the trait so that we know it's intent is to be used only for conversation objects - what do you think?

patshone-gsl (changes requested):
I think we should look to create a specific trait for the conversation object and then a similar and related trait for the regular eloquent models (like User)

patshone-gsl (approved):
Tested locally, this works well

PR #682, opened 2022-11-22
OPNDLG-1742 | Reflection API update to provide boilerplate configuration

elliotmassen:
This PR adds a new optional property to action and interpreter component reflection data. This property allows for default configuration data to be provided for each component, which can be used as a starting point when creating a new configuration for the component.


src/InterpreterEngine/Interpreters/OpenDialogInterpreter.php
```diff
@@ -208,4 +208,17 @@ public function interpreterCallback($intentName, UtteranceAttribute $utterance):
...
         return $intent;
     }
+
+    /**
+     * @inheritDoc
+     */
+    public static function getDefaultConfigurationData(): ?array
+    {
+        return [
+            OpenDialogInterpreterConfiguration::CALLBACKS => [
```
goranambar, line 218: This is not a suggestion, but a question really: why do we need the default callback here?

goranambar (approved):
This looks good Elliot.

I was wondering, do we need to update the other default interpreters also, like `QnaInterpreter` with default configuration data?

src/InterpreterEngine/Interpreters/OpenDialogInterpreter.php
elliotmassen: Right now we actually only require it for actions - which is why I haven't taken the time to create defaults for all interpreters. Including this was more so for demonstrating the pattern in case we make use of this functionality in the future for interpreters.

PR #683, opened 2022-11-23
Handle case when users have the underlying scenario they are interact…

Lintume:
…ing with changed.
The decision requires CORE WebchatInitController, from where the history of the chat is loaded. 
The thing is, to history endpoint has not passed the current scenario id (or alias) (only passed user_id) so we have no ability to understand, whether should we load clear history if the scenario was changed for this user or old history.
I guess we need to change the frontend code, it needs to send a scenario too for a correct understanding of what the content of history needs to be shown.
But I don't know what it can influence.

Related PRs:
[webchat](https://github.com/opendialogai/webchat/pull/398)
[SaaS](https://github.com/opendialogai/opendialog-saas/pull/180)


src/ConversationEngine/Reasoners/ConversationalStateReasoner.php
```diff
@@ -23,12 +24,26 @@ class ConversationalStateReasoner
...
-    public static function determineConversationalStateForUser(UserAttribute $user): ConversationalState
-    {
+    public static function determineConversationalStateForUser(
+        UserAttribute      $user,
+        UtteranceAttribute $utterance = null
+    ): ConversationalState {
         if ($user->hasUserAttribute(UserAttribute::USER_HISTORY_RECORD)) {
+            if ($utterance) {
+                $previousScenarioId = $user->getUserHistoryRecord()->getScenarioId();
+                $currentScenarioId = $utterance->getScenarioId();
```
elliotmassen, line 40: As this method is called after we've established the user, and loaded the User Context, I think we'd be fine to replace this with something like the following. This means the new `$utterance` parameter could be removed.

```suggestion
                $currentScenarioId = ContextService::getAttributeValue('selected_scenario', UserContext::getComponentId());
```

elliotmassen (changes requested):
This looks great - and works just as expected! Thanks so much for this 👍 

Would you be able to add two tests to `ConversationalStateReasonerTest`? It would be great to have one to test the output when the scenario ID's match, and another to test when they don't.

src/ConversationEngine/Reasoners/ConversationalStateReasoner.php
Lintume: thanks, fixed

elliotmassen (approved):

PR #685, opened 2022-11-29
Remove DGraph from OpenDialog code base

Lintume:
[Task in JIRA](https://opendialog.atlassian.net/jira/software/projects/OPNDLG/boards/13?selectedIssue=OPNDLG-1611)


composer.json
```diff
@@ -35,8 +35,7 @@
     "barryvdh/laravel-ide-helper": "^2.7",
     "matthewbdaly/artisan-standalone": "^0.0.8",
     "dms/phpunit-arraysubset-asserts": "^0.2.1",
-    "friendsofphp/php-cs-fixer": "^2.15",
-    "opendialogai/dgraph-docker": "21.03.0.1"
+    "friendsofphp/php-cs-fixer": "^2.15"
   },
   "autoload": {
     "psr-4": {
```
elliotmassen, line 41: Let's remove the `OpenDialogAi\\GraphQLClient\\` mapping from here.

elliotmassen (changes requested):
This looks to be working as expected 👍  Just a few comments below:

 - [x] It looks like there is still a test (`testWebchatHistoryEndpointLimit`) that makes mention to DGraph via a `@requires DGRAPH` method doc comment. Would you be able to remove this?
 - [x] The `ScenarioDataClientInterface.formatScenarioData` method doc comment mentions its use with DGraph, but I can't see where this is used - do you think this should also be removed?
 - [ ] It looks like there's still a few imports/uses of the now removed classes (such as `GraphAttributeDataClient` referenced in `ContextEngineServiceProvider`) - would you be able to remove these?
 - [ ] Should `UserNotPersistedInGraphException` also be removed?

Lintume:
Hey Elliot! thanks a lot for your remarks, agree with all, fixed all.

elliotmassen (approved):
Thanks for making those changes Lialia 👍 I've tested this locally, and it is all working as expected. Nice one!

PR #690, opened 2022-12-16
ODTR-23 | Scenario supported languages

goranambar:
Adds support for `base language` and `supported languages` for `Scenario`. Updates the data clients to store and retrieve the languages properties of the `Scenario`


src/Conversation/Scenario.php
```diff
@@ -147,4 +155,44 @@ public function deactivate(): bool
...
+    {
+        $this->baseLanguage = $value;
+    }
+
+    /**
+     * Gets a list of scenario supported language codes.
+     *
+     * @return ?array
+     */
+    public function getSupportedLanguages(): ?array
```
stu-greenshoots, line 184: I guess this would always return an array as it is initialised?

src/CoreServiceProvider.php
```diff
@@ -64,11 +68,15 @@ public function boot()
...
-                ComponentSettings::class,
-            ]);
-        }
+        $this->commands([
+            ComponentSettings::class,
+            LanguagesSeed::class,
+            ScenariosLanguageUpdate::class
+        ]);
+
+        $this->app->singleton(TranslateServiceInterface::class, function () {
```
stu-greenshoots, line 77: A small change, but could we create a TranslateServiceProvider as per the other modules? Just to keep everything together. 
Eg the ActionEngineServiceProvider

stu-greenshoots (changes requested):
Thanks Goran, I tested this and it looks good to me. Just a couple of very small comments for you to look at

src/Conversation/Scenario.php
goranambar: Hey @stu-greenshoots this is simillar as `status` and `active`, which are both `nullable` and can be set to null in  [dehydrate](https://github.com/opendialogai/core/blob/14cf399f6173ab66f3e4adcfa6b5b684fb83483d/src/Conversation/Scenario.php#L48).

stu-greenshoots (approved):

PR #691, opened 2022-12-19
Feature/conversation engine functional tests

Lintume:
First rude stage of writing tests. Tests are workable but look like noodles.

[Task in Jira](https://opendialog.atlassian.net/jira/software/projects/OPNDLG/boards/13?assignee=6318527c09bc6014ea8c8e37&selectedIssue=OPNDLG-1775)


Tests/Feature/ConversationEngineFunctionalTest.php
```diff
@@ -0,0 +1,1162 @@
...
+            ->setIntentId(PathSubstitutionHelper::findObject($scenario, '$path:webchat/welcome_conversation/welcome_scene/welcome_turn/_requests/0')->getUid())
+            ->setMessageTemplateId(PathSubstitutionHelper::findObject($scenario, '$path:webchat/welcome_conversation/welcome_scene/welcome_turn/_requests/0/0')->getUid())
+            ->setConversationSystem(null);
+
+        $this->assertEquals($expectedPostMessageTemplateMatchState, $postMessageTemplateMatchState);
+
+        // $firstMessage = Hi! This is the default welcome message for the Webchat Scenario.
+
+        /**
+         * 2. Given a intent.core.welcome (callback / button response) utterance, for an existing user,
```
elliotmassen, line 189: I think it will be worth exploring how we can chunk each of these cases up into a separate test method as part of this first phase. That way we have created a way to test from any position in the scenario without having to have previously run the conversation engine. I think this should be possible by artificially giving a user a "user history record" of the desired state.

Tests/Feature/ConversationEngineFunctionalTest.php
```diff
@@ -0,0 +1,1162 @@
...
+        $firstMessage = $messages->first();
+
+        $this->assertNotEquals(MessageTemplateMatcher::ERROR_MESSAGE_ID, $firstMessage->getOdId());
+
+        $this->assertEquals('intent.app.welcomeResponseForWebchat', $firstMessage->getIntent()->getOdId());
+
+        $this->assertEquals(true, ContextService::getAttribute(
+            ActionResult::ACTION_SUCCESS,
+            UserContext::USER_CONTEXT
+        )->getValue());
```
elliotmassen, line 1104: Great idea 👍

Tests/Feature/ConversationEngineFunctionalTest.php
```diff
@@ -0,0 +1,1162 @@
...
+            ->setConversationId(PathSubstitutionHelper::findObject($scenario, '$path:webchat/welcome_conversation')->getUid())
+            ->setSceneId(PathSubstitutionHelper::findObject($scenario, '$path:webchat/welcome_conversation/welcome_scene')->getUid())
+            ->setTurnId(PathSubstitutionHelper::findObject($scenario, '$path:webchat/welcome_conversation/welcome_scene/welcome_turn')->getUid())
+            ->setIntentId(PathSubstitutionHelper::findObject($scenario, '$path:webchat/welcome_conversation/welcome_scene/welcome_turn/_requests/0')->getUid())
+            ->setMessageTemplateId(MessageTemplate::UNDEFINED)
+            ->setConversationSystem(null);
+
+        ConversationalStateReasoner::setConversationalStateForUser($user, $state);
+
+        $user->setRootId('utterance_user');
```
elliotmassen, line 857: It's worth using the constant here instead:

```suggestion
        $user->setRootId(UtteranceAttribute::UTTERANCE_USER);
```

elliotmassen (changes requested):
This is looking good Lialia 🏆  Really great work!

Please see my suggestions below. If possible, I think it would be worth addressing these separately from the optimisation phase. This would mean that we give ourselves the ability to merge this iteration of the tests without blocking ourselves while we iron out the (potentially tricky) work of optimising them for readability / maintainability.

Tests/Feature/ConversationEngineFunctionalTest.php
Lintume: I thought separating every test case independently I'll leave for the next phase, but ok :)

Tests/Feature/ConversationEngineFunctionalTest.php
Lintume: fixed

elliotmassen (approved):

PR #692, opened 2022-12-20
UPDATED php version for circleci config; UPDATED php version, laravel…

andrii-movadex:
… version, guzzle package, symfony/yaml package, symfony/serializer package, symfony/property-access package, laravel/sanctum package, spatie/laravel-event-sourcing package, phpunit/phpunit package, orchestra/testbench package, friendsofphp/php-cs-fixer package; ADDED migration to add event_version field to stored_events table; FIXED normalizers; FIXED MsClientTest tests.


stu-greenshoots (changes requested):
Thanks for this @andrii-movadex , looks good.
1 thing that I'm not too clear on is why our method signatures are different in our supportsNormalization / supportsDenormalization methods. 

I think for clarity, we should have consistent method structures here. Since our methods mostly do in clude the context, I think we should stick to the ContextAware versions of the interface, and add the return type to all methods. Although it's more of a code change, it keeps things consistent for us

elliotmassen (approved):

PR #693, opened 2023-01-03
OPNDLG-1777 conversation engine functional tests improving

Lintume:
[Task in Jira](https://opendialog.atlassian.net/jira/software/projects/OPNDLG/boards/13?assignee=6318527c09bc6014ea8c8e37&selectedIssue=OPNDLG-1777)


Tests/Feature/ConversationEngineFunctionalTest.php
```diff
@@ -2,1231 +2,354 @@
...
+    /**
+     * State in welcome conversation after welcome message from chat
+     *
+     * @param ConversationEngineTestBuilder $scenarioTestBuilder
+     * @return ConversationalState
+     */
+    public function getStateAfterWelcomeMessage(ConversationEngineTestBuilder $scenarioTestBuilder): ConversationalState
+    {
+        if (!$scenarioTestBuilder->scenario) {
+            $this->fail('Be sure that you run mandatory initializeScenario() for $scenarioTestBuilder object');
```
elliotmassen, line 325: Should this use `$this->testCase->fail` instead?

Tests/Utils/Builders/ConversationEngineTestBuilder.php
```diff
@@ -0,0 +1,398 @@
...
+use OpenDialogAi\Core\Conversation\IntentCollection;
+use OpenDialogAi\Core\Conversation\MessageTemplateCollection;
+use OpenDialogAi\Core\Conversation\Scenario;
+use OpenDialogAi\Core\ImportExportHelpers\PathSubstitutionHelper;
+use OpenDialogAi\Core\ImportExportHelpers\ScenarioImportExportHelper;
+use OpenDialogAi\Core\InterpreterEngine\Service\InterpreterServiceInterface;
+use OpenDialogAi\Core\Tests\TestCase;
+use OpenDialogAi\Core\Tests\Utils\UtteranceGenerator;
+
+class ConversationEngineTestBuilder extends TestCase
```
elliotmassen, line 34: Does this need to extend `TestCase` if we use have the `$this->testCase` property?

Tests/Utils/Builders/ConversationEngineTestBuilder.php
```diff
@@ -0,0 +1,398 @@
...
+        return $this;
+    }
+
+    /**
+     * Set up Utterance (what user said)
+     *
+     * @param string|UtteranceAttribute $utteranceOrText
+     * @return $this
+     */
+    public function userSaid($utteranceOrText): ConversationEngineTestBuilder
```
elliotmassen, line 153: What do you think about how to name this method? I'm wondering if they'd be best in present tense, like follows.

```suggestion
    public function userSays($utteranceOrText): ConversationEngineTestBuilder
```

Tests/Utils/Builders/ConversationEngineTestBuilder.php
```diff
@@ -0,0 +1,398 @@
...
+    public ?Scenario $scenario;
+    public ?Map $states;
+    public ?UtteranceAttribute $utterance;
+    public ?UserAttribute $user;
+    public ?array $interpretersToInterpretedIntents;
+    public ?MessageTemplateCollection $messages;
+    public ?ConversationalState $userState;
+
+    public ?string $questionIntentOdId;
+    public ?string $answerIntentOdId;
```
elliotmassen, line 47: I think it'd be worth making these properties protected, so that we have to be intentional about which ones we want to allow for getting/setting.

Tests/Utils/Builders/ConversationEngineTestBuilder.php
```diff
@@ -0,0 +1,398 @@
...
+
+        return $this;
+    }
+
+    /**
+     * Should be called last in chain of functions from this builder. Running chat.
+     *
+     * @return $this|void
+     */
+    public function runConversationEngine()
```
elliotmassen, line 373: Let's rename this method to simply be: `run`. Also, would you be able to add the return type to the method signature please?

Tests/Utils/Builders/ConversationEngineTestBuilder.php
```diff
@@ -0,0 +1,398 @@
...
+        return $this;
+    }
+
+    /**
+     * Setting up user state
+     *
+     * @param ConversationalState $state
+     * @return $this
+     */
+    public function withUserState(ConversationalState $state): ConversationEngineTestBuilder
```
elliotmassen, line 182: What do you think about renaming this to the following?

```suggestion
    public function withConversationalState(ConversationalState $state): ConversationEngineTestBuilder
```

Tests/Utils/Builders/ScenarioAssertionsBuilder.php
```diff
@@ -0,0 +1,204 @@
...
+use OpenDialogAi\Core\ImportExportHelpers\PathSubstitutionHelper;
+use OpenDialogAi\Core\Tests\TestCase;
+
+class ScenarioAssertionsBuilder
+{
+    public TestCase $testCase;
+
+    public bool $isAssertStates = false;
+
+    public function __construct(TestCase $testCase)
```
elliotmassen, line 24: Would this class work is `$testCase` was type `TestCase`? It looks as though perhaps we want to be stricter with the parameter type to be `ConversationEngineTestBuilder`.

Tests/Utils/Builders/ScenarioAssertionsBuilder.php
```diff
@@ -0,0 +1,204 @@
...
+        return $this;
+    }
+
+    /**
+     * @return $this
+     */
+    public function assertNoErrors(): ScenarioAssertionsBuilder
+    {
+        foreach ($this->testCase->messages as $message) {
+            $this->testCase->assertNotEquals(MessageTemplateMatcher::ERROR_MESSAGE_ID, $message->getOdId());
```
elliotmassen, line 45: Let's add an assertion message here, so that if the assertion fails that it's clear what's happened.

Tests/Utils/Builders/ScenarioAssertionsBuilder.php
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        return $this;
+    }
+
+    /**
+     * @param string $intentOdId
+     * @return $this
+     */
+    public function assertChatAnswer(string $intentOdId): ScenarioAssertionsBuilder
```
elliotmassen, line 55: I think we should rename this as follows:

```suggestion
    public function assertResponseContainsMessage(string $intentOdId): ScenarioAssertionsBuilder
```

elliotmassen (changes requested):
This is really great! It's so nice to see how easy to read those tests are now 👍 

Please see my comments below. The majority are centred around ensuring that the naming of the public methods is clear for others who use these classes.

Tests/Utils/Builders/ConversationEngineTestBuilder.php
```diff
@@ -0,0 +1,398 @@
...
+    }
+
+    /**
+     * Resetting Conversation Engine for cleaning chat history
+     *
+     * @return $this
+     */
+    private function renewConversationEngine(): ConversationEngineTestBuilder
+    {
+        $this->conversationEngine = resolve(ConversationEngineInterface::class);
```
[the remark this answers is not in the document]
Lintume: It creates new instance without forgetInstance, it's enough to resolve :)

Tests/Utils/Builders/ScenarioAssertionsBuilder.php
```diff
@@ -0,0 +1,204 @@
...
+use OpenDialogAi\Core\Conversation\Events\ConversationalState\OutgoingIntentStateUpdate;
+use OpenDialogAi\Core\Conversation\Facades\ConversationDataClient;
+use OpenDialogAi\Core\ImportExportHelpers\PathSubstitutionHelper;
+use OpenDialogAi\Core\Tests\TestCase;
+
+class ScenarioAssertionsBuilder
+{
+    public TestCase $testCase;
+
+    public bool $isAssertStates = false;
```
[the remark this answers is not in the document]
Lintume: no, I'll change it to private

Tests/Utils/Builders/ScenarioAssertionsBuilder.php
```diff
@@ -0,0 +1,204 @@
...
+    public function assertChatAnswer(string $intentOdId): ScenarioAssertionsBuilder
+    {
+        $firstMessage = $this->testCase->messages->first();
+        $this->testCase->assertEquals($intentOdId, $firstMessage->getIntent()->getOdId());
+
+        $this->testCase->answerIntentOdId = $intentOdId;
+
+        if ($this->isAssertStates) {
+            $this->assertStates();
+        }
```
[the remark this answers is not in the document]
Lintume: `assertStates()` it's the shortest way to check all states. For checking outcome state is required chat answer intent, that is argument in `assertChatAnswer(string $intentOdId)`. I added the flag `isAssertStates` for the ability to add assertions in any order. so, if `assertChatAnswer()` is added to the chain of assertions after `assertStates()` it will perform assertion states directly in the `assertChatAnswer` function. any suggestions to make it clearer are welcome :)

Tests/Feature/ConversationEngineFunctionalTest.php
Lintume: no, as soon as [ConversationEngineFunctionalTest.php](https://github.com/opendialogai/core/pull/693/files/e4b6325ca75997d24f58a258638c2261e650d5e2#diff-3ccd99e2ebf14efe93de2af102a9f02586b1b27e0ea6e100aadf2cb9e8358c84) extends TestCase we no need to call it by testCase property, Moreover, we have no such property in this class (we no need it)

Tests/Utils/Builders/ConversationEngineTestBuilder.php
```diff
@@ -0,0 +1,398 @@
...
+     * @param array|string $interpretedIntents
+     *
+     * What conversation Engine should hear in the form of intention (optionally with specifying interpreter)
+     * $interpretedIntents can be string with odId or array with format:
+     * [
+     *      name_of_interpreter => odId_of_interpreted_intent
+     * ]
+     * @return $this
+     */
+    public function chatShouldHearIntent($interpretedIntents): ConversationEngineTestBuilder
```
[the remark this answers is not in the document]
Lintume: yes, agree, looks better :)

Tests/Utils/Builders/ConversationEngineTestBuilder.php
Lintume: I thought about setters/getters for protected properties, I'll add them

Tests/Utils/Builders/ConversationEngineTestBuilder.php
Lintume: `run` function already used in TestCase class.
also, this method can return $this or void, but in PHP 7.4 this syntax is not allowed

Tests/Utils/Builders/ConversationEngineTestBuilder.php
Lintume: hmmm... This new name has not reflected the condition of the 'existing user' in the test case. For example, 
"GIVEN Existing user with state after welcome message..."
Here we have the condition of the existing user. I understand that existing user are defined by conversation state, but from the point of intelligibility... what do you think?

Tests/Utils/Builders/ScenarioAssertionsBuilder.php
Lintume: yes, agree, thanks

Tests/Utils/Builders/ScenarioAssertionsBuilder.php
Lintume: oki

Tests/Utils/Builders/ScenarioAssertionsBuilder.php
Lintume: oki :)

Tests/Utils/Builders/ScenarioAssertionsBuilder.php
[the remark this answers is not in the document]
elliotmassen: Ah I see, thanks for clarifying! I think it's fine as it for now 👍

Tests/Utils/Builders/ConversationEngineTestBuilder.php
elliotmassen: Ah good point - my bad! No worries, let's leave it as is 👍

Tests/Utils/Builders/ConversationEngineTestBuilder.php
elliotmassen: That's a fair point - I get what you mean. However, my thinking was that "with user state" potentially implies that there is some other kind of state (i.e. "with app state"). Currently there is only conversational state that is owned by a user (in user history record), however these state objects can also represent states relative to both the user and the application. I think it might be confusing for someone to see a `$state` with a speaker of `Intent::APP`, but then to see that state passed to a `withUserState` method.

Let me know what you think!

src/Conversation/DataClients/Relational/ConversationDataClient.php
```diff
@@ -714,6 +714,47 @@ public function getIntentByUid(string $intentUid): Intent
         return $intent;
     }
 
+    public function getIntentByOdId(string $intentOdId): Intent
+    {
+        $intentModel = ConversationObjectBase::where('od_id', $intentOdId)->first();
```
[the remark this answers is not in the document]
Lintume: yes, agree, I'll add checking type and scenario

src/Conversation/DataClients/Relational/ConversationDataClient.php
```diff
@@ -714,6 +714,47 @@ public function getIntentByUid(string $intentUid): Intent
         return $intent;
     }
 
+    public function getIntentByOdId(string $intentOdId): Intent
```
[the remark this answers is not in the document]
Lintume: totally agree :)

Tests/Feature/ConversationEngineFunctionalTest.php
```diff
@@ -289,47 +209,148 @@ public function testWelcomeTextUtteranceWithDifferentInterpreter()
...
+            ->shouldNotPerformActions()
+            ->runConversationEngine();
 
-            ->assertMessageTemplateState($this->getStateAfterWelcomeMessage($scenarioTestBuilder)
-                ->setMessageTemplateId($scenarioTestBuilder->getUidFromPath('$path:webchat/welcome_conversation/welcome_scene/welcome_turn/_requests/0/0')))
+        $assertions = $this->initializeAssertions($scenarioTestBuilder);
+        $assertions->assertGotMessages(2)
+            ->assertNoErrors()
+            ->assertResponseContainsMessages('questionPreambleResponse', 'helpResponse')
+            ->assertIncomingIntentState($scenarioTestBuilder->getStateFromIntent('helpRequest'))
```
elliotmassen, line 275: Is there anyway we are able to make an assertion about both incoming states (the one before virtual intent, and the one after)? If so, it'd be worth looking to do the same for the outgoing and message template states too.

Tests/Feature/ConversationEngineFunctionalTest.php
```diff
@@ -289,47 +209,148 @@ public function testWelcomeTextUtteranceWithDifferentInterpreter()
...
+            ->runConversationEngine();
+
+        $assertions = $this->initializeAssertions($scenarioTestBuilder);
+        $assertions->assertGotOneMessage()
+            ->assertNoErrors()
+            ->assertResponseContainsMessage('handHandoverResponse')
+            ->assertIncomingIntentState()
+            ->assertOutgoingIntentState()
+            ->assertMessageTemplateState(ConversationContextUtil::createAsConversationalState()
+                ->setConversationSystem(null)) // Event of changing Message Template State firing before Conversation System changing
```
elliotmassen, line 308: Ah that's a good spot - I didn't realise this!

It makes me wonder if we should update the `ConversationEngine.processMessageTemplate` method so that the state is manipulated before we fire the event. That way it would be consistent with the `ConversationEngine.updateState` method, where the state is created from the intent (including any transition / completing behaviour manipulations). What do you think? Perhaps we would need to consider if that might break any existing functionality.

Tests/Feature/ConversationEngineFunctionalTest.php
```diff
@@ -289,47 +209,148 @@ public function testWelcomeTextUtteranceWithDifferentInterpreter()
...
+        // With the help of this mechanism we make the transition from 'intent.core.TurnNoMatch' intent to 'moreInfoResponse'
+        // instead of expected response 'intent.app.welcomeResponseForWebchat'.
+
+        $state = $scenarioTestBuilder->getStateFromIntent('moreInfoRequest');
+        $actionResult = ActionResult::createSuccessfulActionResultWithAttributes([
+            new StringAttribute('current_conversation', $state->getConversationId()),
+            new StringAttribute('current_scene', $state->getSceneId()),
+            new StringAttribute('current_turn', $state->getTurnId()),
+            new StringAttribute('turn_status', $state->getTurnStatus())
+        ]);
```
elliotmassen, line 341: It's worth noting that this slightly different from what we had discussed, however it does still test dynamic transitioning - so it's not incorrect at all!

The state that is being generated from that `moreInfoRequest` creates a `MID_TURN` state. This means that when the action is run, we update the conversational position to the middle of that turn (as if we had just matched `moreInfoRequest` intent). We haven't previously used dynamic transitions like this, but it's very cool to see that this is possible!

In order to keep the test more straightforward, I think it would be worth using a state for the Welcome turn, with the status of `IN_TURN`. This would model a dynamic (turn) transition. The Welcome turn has an application request intent, so the cadence is already as you required. You would be able to easily get this state by using `getStateFromIntent` on the `intent.core.welcome` intent in the Trigger conversation I think.

elliotmassen (changes requested):
Thanks for making those changes, and including the complex tests. I think the comments above each test case are super useful, thank you for including them 👍  Just a few further comments.

src/Conversation/DataClients/Relational/IntentDataClient.php
```diff
@@ -162,4 +159,54 @@ public function getAllIntentsInScenario(string $scenarioId)
 
         return $intentCollection;
     }
+
+    public function getIntentByOdId(string $intentOdId, string $scenarioUid): IntentCollection
```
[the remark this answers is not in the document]
Lintume: ah, sorry, my bad, thanks, I'll add it

Tests/Feature/ConversationEngineFunctionalTest.php
Lintume: I've added ability to check all states changing.

Tests/Feature/ConversationEngineFunctionalTest.php
Lintume: I've changed `ConversationEngine.processMessageTemplate` in a way that the template status updating after the conversation system change

Tests/Feature/ConversationEngineFunctionalTest.php
Lintume: I hope I've done this :)

elliotmassen (approved):
This is really great! 🏆 Thanks for making these changes - I'm happy for this to be merged.

PR #695, opened 2023-01-09
Adds translatable property to the component settings table

stu-greenshoots:
This PR adds a new 'translatable' property to the component settings table to mark which settings are translatable. This is used in this webchat PR https://github.com/opendialogai/webchat/pull/404.

A new scope has been added to fetch only the translatable strings for easy of use


goranambar (approved):
Ok, I tested it by running the migration in `saas` and together with other PR's, from `webchat` and  `od-design-system` I can see that it works as expected.

PR #696, opened 2023-01-09
ODTR-16 | New way of handling XML messages

goranambar:
Add `fromXml` method for creating messages directly from `xml`.


stu-greenshoots (approved):
This looks good and has good test coverage. We already know that the options message is missing, but will need to be added when the options message PR is merged.

Happy for this to be merged now - if there are any issues with particular message properties not working as expected, these can be picked up at the time in the appropriate translate ticket

PR #699, opened 2023-01-16
Options Message type

stu-greenshoots:
This PR adds a new message type for 'options message'. There is a partial functionality PR already merged in webchat here https://github.com/opendialogai/webchat/pull/405

This back-end PR adds full support for the message type in the response engine and the webchat message formatter. The message is expected to be in the format:

```
<options-message>
    <external>$this->external</external>
    <submit_text>$this->submitText</submit_text>
    <callback>$this->callback</callback>
    <cancel_text>$this->cancelText</cancel_text>
    <cancel_callback>$this->cancelCallback</cancel_callback>
    <options>
        <option type="$this->type">
            <attribute>$this->attribute</attribute>
            <label>$this->label</label>
            <required>$this->required</required>
        </option>
    </options>
</options-message>
```

The cancel callback and cancel text are optional if a cancel button is to be shown.

Each `option` provides the attribute associated with it, a label and whether it is required. The intention is that if an option is required, it needs to be selected for the message to be submittable. Only `checkbox` options are expected to be supported for now.

The `external` property is to indicate whether the buttons should be external to the message in the same way as external button messages


src/MessageBuilder/Message/OptionsMessage.php
```diff
@@ -0,0 +1,63 @@
...
+    {
+        $this->options[] = $option;
+    }
+
+    public function getOptionsMarkup()
+    {
+        $optionsMarkUp = '';
+
+        foreach ($this->options as $option) {
+            $optionsMarkUp = $option->getMarkUp();
```
goranambar, line 40: Shouldn't this be `.=` instead of just `=`?

goranambar (changes requested):
@stu-greenshoots this looks OK. Just the `markup` thing I've mentioned in my comment.

src/MessageBuilder/Message/OptionsMessage.php
stu-greenshoots: yeah, good spot!

PR #704, opened 2023-01-31
Adds ComponentConfigurationTranslation Helper

goranambar:
Adds ComponentConfiguration Translation helper that has methods to put and pull configuration translation strings from and to translation table.


src/Translate/Helpers/ComponentConfigurationTranslationHelper.php
```diff
@@ -0,0 +1,77 @@
...
+                        $setting->name,
+                        $componentConfiguration->scenario_id,
+                        $value,
+                        $language
+                    );
+                    $configurations[$configurationName][$key] = $translateId;
+                }
+            }
+        }
+        $componentConfiguration->configuration = $configurations;
```
oliversarfas, line 43:
```suggestion
        $componentConfiguration->configuration = $configurations;
        $componentConfiguration->save();
```

We're currently not persisting the config changes, `save()` will make this happen

src/Translate/Helpers/ComponentConfigurationTranslationHelper.php
```diff
@@ -0,0 +1,77 @@
...
+                    $translationString = TranslateService::getConfigTranslationString(
+                        $setting->name,
+                        $componentConfiguration->scenario_id,
+                        $language
+                    );
+                    $configurations[$configurationName][$key] = $translationString;
+                }
+            }
+        }
+        $componentConfiguration->configuration = $configurations;
```
oliversarfas, line 75: Does this need `save()` also?

src/Translate/Helpers/ComponentConfigurationTranslationHelper.php
goranambar: No, this should not be the responsibility of the helper. Helper just pulls the translatable strings, adds them into translation table and updates the translatable string with ids on the ComponentConfiguration. After that, the calling code can decide wether to save the configuration or not.

oliversarfas (approved):

src/Translate/Helpers/ComponentConfigurationTranslationHelper.php
oliversarfas: For this to work the $componentConfiguration needs to be passed by reference, or returned

goranambar: Hey @oliversarfas, that works, because in php objects are passed by reference by default. [Here's](https://onlinephp.io/c/75775) a quick example.

oliversarfas (approved):

oliversarfas (approved):

PR #706, opened 2023-02-03
ADDED laravel/scout package; UPDATED composer.json and composer.lock …

andrii-movadex:
…files; CHANGED ConversationObjectBase model: added factory and searchable traits; ADDED query scopes for search; ADDED return of searchable array.


elliotmassen (approved):
This looks good to me @andrii-movadex. Before this is ready to merge, would you be able to confirm for me if the `composer.lock` file changes we due to installing Scout alone, or did you also run `composer update` at some point?

andrii-movadex:
> This looks good to me @andrii-movadex. Before this is ready to merge, would you be able to confirm for me if the `composer.lock` file changes we due to installing Scout alone, or did you also run `composer update` at some point?

In this case I've just removed composer.lock and re-created it via composer install (as far as I remember).

elliotmassen:
> > This looks good to me @andrii-movadex. Before this is ready to merge, would you be able to confirm for me if the `composer.lock` file changes we due to installing Scout alone, or did you also run `composer update` at some point?
> 
> In this case I've just removed composer.lock and re-created it via composer install (as far as I remember).

Thanks for clarifying! I've just updated the lock file to reduce the amount of changes in this case. We would usually prefer to be more intentional with updating dependencies, so for a PR like this, we'd just want to keep the current lock file and run `composer require/update {dependency}`.

andrii-movadex:
> > > This looks good to me @andrii-movadex. Before this is ready to merge, would you be able to confirm for me if the `composer.lock` file changes we due to installing Scout alone, or did you also run `composer update` at some point?
> > 
> > 
> > In this case I've just removed composer.lock and re-created it via composer install (as far as I remember).
> 
> Thanks for clarifying! I've just updated the lock file to reduce the amount of changes in this case. We would usually prefer to be more intentional with updating dependencies, so for a PR like this, we'd just want to keep the current lock file and run `composer require/update {dependency}`.

ok, got it. Thanks

PR #708, opened 2023-02-07
ODTR 56 | Add message translation helper

goranambar:
Adds message translation helper that strips all of the translatable content from Template messages and adds it to the translation table.


src/CoreServiceProvider.php
```diff
@@ -69,7 +70,8 @@ public function boot()
         $this->commands([
             ComponentSettings::class,
             LanguagesSeed::class,
-            ScenariosLanguageUpdate::class
+            ScenariosLanguageUpdate::class,
+            StripScenariosTranslatableStrings::class
```
oliversarfas, line 74: Minor thing, add trailing comma to make next update cleaner in git diff

```suggestion
            StripScenariosTranslatableStrings::class,
```

src/Translate/Helpers/ComponentConfigurationTranslationHelper.php
```diff
@@ -0,0 +1,151 @@
...
+                return TranslateService::getConfigTranslationString(
+                    $setting->name,
+                    $scenarioId,
+                    $language
+                );
+            } catch (TranslationNotFoundException $e) {
+                Log::warning(sprintf('Translation for configuration: %s, for language: %s ' .
+                    'for scenario: %s not found.', $setting->name, $language, $scenarioId));
+                return $value;
+                // TODO: should we revert to the base language here, if there's no translation with the
```
oliversarfas, line 129: I prefer the way you've got it already. If a user asks for translation in 'es' but we don't have it, return their original value. If we defaulted to 'jp' or something, it'd be confusing

src/Translate/Helpers/MessageTemplateTranslationHelper.php
```diff
@@ -0,0 +1,495 @@
...
+                    $messageCount++;
+                    break;
+                case OpenDialogMessage::DATE_PICKER_MESSAGE:
+                    self::addDatePickerMessageTranslation($uid, $messageCount, $message, $scenarioId, $language);
+                    $messageCount++;
+                    break;
+                case OpenDialogMessage::AUTOCOMPLETE_MESSAGE:
+                    self::addAutoCompleteMessageTranslation($uid, $messageCount, $message, $scenarioId, $language);
+                    $messageCount++;
+                    break;
```
oliversarfas, line 94: Can we run $messageCount++ after the switch? Seems unnecessary to have it in each case

oliversarfas (changes requested):
Great PR @goranambar, looks great and nice test coverage.

Could we add a test for when the translation is not found that we return the original value requested? Doesn't have to be for each message type, just a basic proof of concept that it works should suffice as we have tests for the rest already

src/Translate/Helpers/ComponentConfigurationTranslationHelper.php
goranambar: Falling back to base language is a requirement by design, @oliversarfas . Just not sure at the moment if we should do it in this place.

src/Translate/Helpers/MessageTemplateTranslationHelper.php
goranambar: @oliversarfas not all messages have translatable content and we want to count only the ones that do.

goranambar:
Thanks @oliversarfas . Those tests you required were actually from [previous approved PR](https://github.com/opendialogai/core/pull/704) :) But I have added them here, nonetheless. Sorry for the confusion.

oliversarfas (approved):

PR #709, opened 2023-02-10
Adds forceUpdate to ComponentConfigurationTranslationHelper

oliversarfas:
add boolean to allow for bypas of update and return existing string instead


goranambar (changes requested):
Hey @oliversarfas I'm afraid this is not what we're looking for. I dropped some directions on how to implement this.

src/Translate/Services/TranslateService.php
```diff
@@ -163,7 +172,9 @@ protected static function addOrUpdateLanguageLine($group, $key, $type, $scenario
             );
         }
 
-        $languageLine->setTranslation($language, $translationString);
+        if ($forceUpdate) {
```
goranambar, line 175: So, if `$forceUpdate` is set to `false`, and `LanguageLine` didn't exist previously (as it is created few lines before), it won't set the translation string to it. And we don't want that, right?

src/Translate/Services/TranslateService.php
```diff
@@ -146,10 +147,18 @@ protected static function generateMessageKey($uid, $name): string
...
+        $group,
+        $key,
+        $type,
+        $scenarioId,
+        $language,
+        $translationString,
+        bool $forceUpdate = true,
+    ): string {
         $languageLine = OpenDialogLanguageLine::where('group', $group)->where('key', $key)->first();
         if (!$languageLine) {
```
goranambar, line 163: Just check before this line:
`if ($languageLine && !$forceUpdate){return;}`
And delete the `$forceUpdate` check below.

goranambar (changes requested):
Still couple of points to fix.

src/Translate/Services/TranslateService.php
```diff
@@ -86,7 +86,8 @@ public function addOrUpdateCustomTranslation(
         string $name,
         int $scenarioId,
         string $translationString,
-        string $language
+        string $language,
+        bool $forceUpdate,
```
goranambar, line 90: This should be not in here but in `addOrUpdateConfigTranslation` right?

src/Translate/Services/TranslateServiceInterface.php
```diff
@@ -103,7 +103,8 @@ public function addOrUpdateCustomTranslation(
         string $name,
         int $scenarioId,
         string $translationString,
-        string $language
+        string $language,
+        bool $forceUpdate,
```
goranambar, line 107: This should be not in here but in addOrUpdateConfigTranslation right?

oliversarfas:
@goranambar test added and PR feedback actioned

goranambar (approved):
Looks OK now

PR #711, opened 2023-02-13
ODTR 41 | Generate CSV translation export file

goranambar:
Adds function in `TranslateService` for generation of translation files for export and implements Csv exporter and adds `scenario:export-translation` command.


src/Translate/Tests/TranslateServiceTest.php
```diff
@@ -102,4 +103,23 @@ public function testDeleteTranslationsByScenario()
...
+        TranslateService::addOrUpdateMessageTranslation(12345, 'button', 'button-message', 123, 'some string in en', 'en');
+        TranslateService::addOrUpdateConfigTranslation('title', 123, 'some other string, in en', 'en');
+        $this->assertEquals(2, OpenDialogLanguageLine::where('scenario_id', 123)->count());
+
+        $csv = TranslateService::exportTranslationFile(123, 'en', ['es', 'pt'], 'csv');
+        $lines = explode("\n", $csv);
+        $this->assertEquals('group,key,domain,status,en,es,pt', $lines[0]);
+        $this->assertEquals('123-message,12345-button,button-message,New,"some string in en","",""', $lines[1]);
+        $this->assertEquals('123-config,title,config,New,"some other string, in en","",""', $lines[2]);
+    }
```
oliversarfas, line 118: Can we please add a test for the file generation, filename thereof, and that the contents match what we're expecting?

goranambar: This can't be done here. This is not saving to a file, but just generating csv content.

src/Console/Commands/ScenarioExportTranslation.php
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        return 0;
+    }
+
+    private function getExportDir(): string
+    {
+        $dir = storage_path() . DIRECTORY_SEPARATOR . self::EXPORT_DIR;
+        if (!File::isDirectory($dir)) {
+            File::makeDirectory($dir, 0777, true, true);
```
oliversarfas, line 70: Is `0777` the correct permission to be using here? Would `0775` make more sense so that we're not allowing execution for files? I know we're the one generating the file so the risk is minimal, however still a risk

oliversarfas (approved):
Slight comment regarding local file permissions on system, however as we're generating the files ourselves not a deal breaker. LGTM

PR #715, opened 2023-02-17
Implement language processor component service (with interfaces)

andrii-movadex:


src/Components/Helper/ComponentHelper.php
```diff
@@ -14,6 +14,7 @@ class ComponentHelper
 
     const ACTION_ID_PATTERN = "/^action\.[A-z]*\.[A-z_]*$/";
     const INTERPRETER_ID_PATTERN = "/^interpreter\.[A-z]*\.[A-z_]*$/";
+    const NLU_ID_PATTERN = "/^language_service\core*\.[A-z_]*$/";
```
elliotmassen, line 17: Let's update this constant name to be as clear as possible.

```suggestion
    const LANGUAGE_SERVICE_ID_PATTERN = "/^language_service\core*\.[A-z_]*$/";
```

src/NluEngine/Component/NluComponentInterface.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace OpenDialogAi\NluEngine\Component;
```
elliotmassen, line 3: These classes should all live in a new Language Service namespace and directory. The NLU Engine is a different (outdated) engine.

src/NluEngine/Component/NluComponentInterface.php
```diff
@@ -0,0 +1,19 @@
...
+
+use OpenDialogAi\AttributeEngine\CoreAttributes\UtteranceAttribute;
+use OpenDialogAi\InterpreterEngine\Interpreters\AbstractNLUInterpreter\AbstractNLURequestFailedException;
+use OpenDialogAi\NluEngine\Response\ResponseInterface;
+
+interface NluComponentInterface
```
elliotmassen, line 9: Let's rename these classes after "LanguageService" rather than "Nlu".

```suggestion
interface LanguageServiceComponentInterface
```

src/NluEngine/Service/NluComponentService.php
```diff
@@ -0,0 +1,83 @@
...
+    public function register(string $class): void
+    {
+        if (!class_exists($class) || !in_array(NluComponentInterface::class, class_implements($class))) {
+            Log::warning(
+                sprintf("NLU component %s was not registered, as it doesn't implement the correct interface.", $class)
+            );
+            return;
+        }
+
+        $id = $class::getComponentId();
```
elliotmassen, line 62: I think it would be worth adding a call to `$class::getComponentData()` above this line (like the Interpreter Engine) so that we are able to throw any errors related to incorrect set up of component classes (i.e. `MissingRequiredComponentDataException` from unset ID, type or source.

src/NluEngine/Component/NluComponentInterface.php
```diff
@@ -0,0 +1,19 @@
...
+{
+    /**
+     * Interprets an utterance and returns all matching intents in an array
+     *
+     * @param UtteranceAttribute $utterance
+     * @return ResponseInterface
+     * @throws AbstractNLURequestFailedException
+     */
+    public function interpret(UtteranceAttribute $utterance) : ResponseInterface;
```
elliotmassen, line 18: As per the Interpreter component service interface, we will also need a signature for `getConfiguration`, `createConfiguration`, and `getDefaultConfigurationData` in this interface.

src/NluEngine/Service/NluComponentService.php
```diff
@@ -0,0 +1,83 @@
...
+
+use Ds\Map;
+use Illuminate\Support\Facades\Log;
+use OpenDialogAi\Core\Components\Exceptions\ComponentNotRegisteredException;
+use OpenDialogAi\Core\Components\Helper\ComponentHelper;
+
+class NluComponentService implements NluComponentServiceInterface
+{
+    /**
+     * @var Map|InterpreterInterface[]
```
elliotmassen, line 13: Could you correct the doc blocks please?

elliotmassen (changes requested):
- [ ] Would you be able to ensure that these PR's have titles which reflect the changes they include?
- [ ] Let's get some tests added for this service.
- [ ] Please could you ensure that these classes live within a `LanguageServiceEngine` directory and namespace.

src/LanguageServiceEngine/Service/LanguageServiceComponentServiceInterface.php
```diff
@@ -0,0 +1,44 @@
...
+     * @return bool
+     */
+    public function has(string $componentId): bool;
+
+    /**
+     * Registers a language service component (which includes a component ID).
+     *
+     * @param string $class
+     */
+    public function register(string $class): void;
```
elliotmassen, line 43: Let's add a `@throws` comment here to document that this method can throw `MissingRequiredComponentDataException`.

src/LanguageServiceEngine/Component/LanguageServiceComponentInterface.php
```diff
@@ -0,0 +1,41 @@
...
+     * @return array|null
+     */
+    public static function getDefaultConfigurationData(): ?array;
+
+    /**
+     * Interprets an utterance and returns matching response
+     *
+     * @param UtteranceAttribute $utterance
+     * @return ResponseInterface
+     * @throws AbstractNLURequestFailedException
```
elliotmassen, line 38: Should this be here?

elliotmassen (changes requested):
Thanks for making those changes Andrii! Just a few comments below:

- [ ] As requested in the last review, would you be able to ensure that these PR's have titles which reflect the changes they include?
- [ ] Would you be able to update the directory, namespace and classes to be `LanguageProcessor` rather than `LanguageService`? Apologies for the second change to this, however we have taken a decision that all backend references to "Language Services" would be clearer if we refer to them as "Language Processors" - the frontend will still refer to them as Services however (it's simply a way to make it easier for us to talk about them in the backend).

src/LanguageServiceEngine/Configuration/BaseLanguageServiceConfiguration.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace OpenDialogAi\LanguageServiceEngine\Configuration;
+
+use OpenDialogAi\Core\Components\Configuration\BaseConfiguration;
+
+class BaseLanguageServiceConfiguration extends BaseConfiguration implements LanguageServiceConfigurationInterface
```
elliotmassen, line 7: Does this class need to override `getConfigurationKey` or similar in order to account for the fact that the key won't contain a scenario ID? Currently the key's have a non-nullable `scenarioId` property. I think we should explicitly account for this (rather than providing empty string as scenario ID).

src/LanguageServiceEngine/Service/LanguageServiceComponentServiceInterface.php
andrii-movadex: Fixed.

src/LanguageServiceEngine/Component/LanguageServiceComponentInterface.php
andrii-movadex: Fixed.

src/LanguageServiceEngine/Configuration/BaseLanguageServiceConfiguration.php
andrii-movadex: I've added LanguageProcessorConfigurationKey and overrided getConfigurationKey method inside BaseLanguageProcessorConfiguration class. I am not sure this is the best solution but the other way is to update getConfigurationKey inside BaseConfiguration file to accept nullable values also. Please check

src/Components/Helper/ComponentHelper.php
```diff
@@ -14,7 +14,7 @@ class ComponentHelper
 
     const ACTION_ID_PATTERN = "/^action\.[A-z]*\.[A-z_]*$/";
     const INTERPRETER_ID_PATTERN = "/^interpreter\.[A-z]*\.[A-z_]*$/";
-    const LANGUAGE_SERVICE_ID_PATTERN = "/^language_service\.core\.[A-z_]*$/";
+    const LANGUAGE_PROCESSOR_ID_PATTERN = "/^language_processor\.core\.[A-z_]*$/";
```
elliotmassen, line 17: These component ID's can be registered with any scope, not just `core`. Let's make this consistent with the established patterns:

```suggestion
    const LANGUAGE_PROCESSOR_ID_PATTERN = "/^language_processor\.[A-z]*\.[A-z_]*$/";
```

src/LanguageServiceEngine/Configuration/BaseLanguageServiceConfiguration.php
elliotmassen: Thanks Andrii! I think we might need to consider a different solution here. Currently, as the new class extends the existing one, it would still be possible to call `getScenarioId` on an object of the new class, which would cause an error currently as the `scenarioId` property is undefined.

To keep the codebase more maintainable, I think it might be worth introducing a new class like `ComponentConfigurationKeyWithName` that only contains the `configurationName` property (and associated methods), and then to make the existing `ComponentConfigurationKey` class extend this new class. This way we can use the new `ComponentConfigurationKeyWithName` to key the Language Processors. The new inheritance chain would be:

`Hashable` -> `ComponentConfigurationKeyWithName` -> `ComponentConfigurationKey`

elliotmassen (changes requested):
Thanks for making those changes! A few final comments below.

elliotmassen (approved):

PR #716, opened 2023-02-17
Implement configured language processor service (with interfaces)

andrii-movadex:


elliotmassen (changes requested):
Alongside my feedback given in https://github.com/opendialogai/core/pull/715#pullrequestreview-1307842326, could you add some tests for this too please?

I also think it would be worth adding an abstract class similar to `BaseInterpreter`.

src/LanguageServiceEngine/Tests/Helper/HelperToProvideLanguageServices.php
```diff
@@ -0,0 +1,35 @@
...
+            return new LanguageServiceComponentService();
+        });
+    }
+    
+    /**
+     * @param  Application $app
+     * @param  LanguageServiceComponentServiceInterface $languageServiceComponentService
+     * @return void
+     */
+    public function provideConfiguredLanguageServiceService(Application $app, LanguageServiceComponentServiceInterface $languageServiceComponentService)
```
elliotmassen, line 29: Do we need this method? I'm thinking that this should be happening anyway in a service provider.

src/Components/ODComponentTypes.php
```diff
@@ -16,6 +16,7 @@ class ODComponentTypes
     const SENSOR_COMPONENT_TYPE = 'sensor';
     const PLATFORM_COMPONENT_TYPE = 'platform';
     const ATTRIBUTE_FILTER_TYPE = 'attribute_filter';
+    const LANGUAGE_SERVICE_COMPONENT_TYPE = 'language_service';
```
elliotmassen, line 19: Let's get this updated to:

```suggestion
    const LANGUAGE_PROCESSOR_COMPONENT_TYPE = 'language_processor';
```

elliotmassen (changes requested):

src/LanguageServiceEngine/Tests/Service/ConfiguredLanguageServiceServiceTest.php
```diff
@@ -0,0 +1,106 @@
...
+
+class ConfiguredLanguageServiceServiceTest extends TestCase
+{
+    use HelperToProvideLanguageServices;
+    use LanguageServiceComponentTestHelper;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+    }
```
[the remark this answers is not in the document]
andrii-movadex: Yes, removed it.

src/LanguageServiceEngine/Tests/Helper/HelperToProvideLanguageServices.php
andrii-movadex: yes, but there is no service provider now.. And I need somehow resolve bindings. May be I can resolve class instead of interface and it will be working. I am not sure what is the best way here

elliotmassen (approved):
Thanks for making those changes - this is looking good 👍

src/LanguageServiceEngine/Tests/Helper/HelperToProvideLanguageServices.php
andrii-movadex: ok, I will tidy up this when I will setup the provider. Agreed

PR #717, opened 2023-02-20
ODTR-87 | Adds getTranslationLanguagesByScenario function

goranambar:
Adds getTranslationLanguagesByScenario function to the `TranslateService`


src/Translate/Services/TranslateService.php
```diff
@@ -255,4 +256,33 @@ public function deleteConfigTranslationsByScenario(int $scenarioId): void
...
+                $scenarioId,
+                $language->code,
+                $language->name,
+                $language->native_name,
+                $language->rtl,
+                100
+            );
+        }
+
+        return $result;
```
oliversarfas, line 286: anyway we can refactor to use a map instead of variable assignment and loops?

oliversarfas (approved):
Looks good. Just a comment about code style using array foreach over collection->map, but not a deal breaker

PR #718, opened 2023-02-27
Implement language processor service (with interfaces)

andrii-movadex:


src/LanguageServiceEngine/Service/LanguageServiceService.php
```diff
@@ -0,0 +1,49 @@
...
+
+use Exception;
+use Illuminate\Support\Facades\Log;
+use OpenDialogAi\AttributeEngine\CoreAttributes\UtteranceAttribute;
+use OpenDialogAi\Core\Components\Configuration\ComponentConfigurationKey;
+use OpenDialogAi\Core\Components\Exceptions\ConfigurationNotRegistered;
+use OpenDialogAi\LanguageServiceEngine\Response\EmptyResponse;
+use OpenDialogAi\LanguageServiceEngine\Response\ResponseInterface;
+
+class LanguageServiceService implements LanguageServiceServiceInterface
```
elliotmassen, line 13:
```suggestion
class LanguageProcessorService implements LanguageProcessorServiceInterface
```

src/LanguageServiceEngine/Service/LanguageServiceService.php
```diff
@@ -0,0 +1,49 @@
...
+
+    public function __construct(ConfiguredLanguageServiceServiceInterface $configuredLanguageServiceService)
+    {
+        $this->configuredLanguageServiceService = $configuredLanguageServiceService;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function interpret(ComponentConfigurationKey $key, UtteranceAttribute $utterance): ResponseInterface
```
elliotmassen, line 25: This is signature is exactly right, nice one 👍

src/LanguageServiceEngine/Service/LanguageServiceService.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        $name = $key->getConfigurationName();
+
+        try {
+            /** @var LanguageServiceComponentInterface $languageServiceComponent */
+            $languageServiceComponent = $this->configuredLanguageServiceService->get($key);
+            $componentId = $languageServiceComponent::getComponentId();
+        } catch (ConfigurationNotRegistered $e) {
+            Log::error(sprintf("Language service component %s is not a registered.", $name));
+            return new EmptyResponse();
```
elliotmassen, line 35: I think this exception should be thrown by this method, as there is a difference between an empty response and no response.

src/LanguageServiceEngine/Service/LanguageServiceService.php
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * @inheritdoc
+     */
+    public function interpret(ComponentConfigurationKey $key, UtteranceAttribute $utterance): ResponseInterface
+    {
+        $name = $key->getConfigurationName();
+
+        try {
+            /** @var LanguageServiceComponentInterface $languageServiceComponent */
```
elliotmassen, line 30: PhpStorm is saying that we are missing this interface as an import.

elliotmassen (changes requested):
Thanks Andrii! Please see my comments below.

src/LanguageServiceEngine/Response/ResponseInterface.php
```diff
@@ -4,4 +4,8 @@
 
 interface ResponseInterface
 {
+    /**
+     * @return array
+     */
+    public function getData() : array;
```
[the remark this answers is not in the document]
andrii-movadex: Well, I added that because:
- I was thinking I would require to get some data from response and this method could be helpful
- and also to be able to add tests and check if the service working
I've removed this now because it is also possible to work without this method and according to the future response (you've provided me an example of that) there should not be such a method.

src/LanguageServiceEngine/Service/LanguageServiceService.php
elliotmassen: I think this issue is still outstanding, however just in the `LanguageProcessorService` now.

src/LanguageServiceEngine/Service/LanguageServiceService.php
```diff
@@ -0,0 +1,49 @@
...
+            Log::error(sprintf("Language service component %s is not a registered.", $name));
+            return new EmptyResponse();
+        }
+
+        try {
+            $response = $languageServiceComponent->interpret($utterance);
+        } catch (Exception $e) {
+            Log::warning(sprintf('Language service component %s (%s) failed with exception: %s',
+                                $name, $componentId, $e->getMessage()
+                        ));
```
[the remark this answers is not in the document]
elliotmassen: I think this issue is still outstanding, however just in the `LanguageProcessorService` now.

src/LanguageServiceEngine/Tests/Service/LanguageServiceServiceTest.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        // Check if language service's service exist
+        $this->assertTrue(interface_exists(LanguageServiceServiceInterface::class));
+        $this->assertTrue(class_exists(LanguageServiceService::class));
+
+        // Check if LanguageServiceService implements LanguageServiceServiceInterface
+        $this->assertTrue(in_array(LanguageServiceServiceInterface::class, class_implements(LanguageServiceService::class)));
+    }
+
+    public function testLanguageServiceService()
```
elliotmassen: Would you be able to update this test name to the new processor phrasing?

src/LanguageServiceEngine/Tests/Service/LanguageServiceServiceTest.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        // Setup configuration
+        $name = 'example';
+
+        /** @var EmptyLanguageServiceConfiguration $configuration */
+        $configuration = LanguageServiceModelExample::createConfiguration($name)->setConfigurationScenarioId('0x000');
```
elliotmassen: I think this may need updating in line with the changes from the other PR's (re. the scenario ID, and the `LanguageServiceModelExample` class name.

src/LanguageServiceEngine/Tests/Service/LanguageServiceServiceTest.php
```diff
@@ -0,0 +1,89 @@
...
+
+class LanguageServiceServiceTest extends TestCase
+{
+    use HelperToProvideLanguageServices;
+    use LanguageServiceServiceTestHelper;
+
+    protected function setUp(): void
+    {
+       parent::setUp();
+    }
```
elliotmassen: Let's remove this.

elliotmassen (changes requested):

src/LanguageServiceEngine/Service/LanguageServiceService.php
andrii-movadex: Sorry, I am not sure I clearly understood this. So should I remove all the catch blocks?

elliotmassen (approved):

PR #719, opened 2023-03-01
ODTR-57 | Update Message Formatter in Response Engine

oliversarfas:
Runs through all messages in response engine and swaps out the necessary parts to return the translation instead of translation key where appropriate

---

Known tech debt / todo

- Pull in target language instead of hard coded `en`.
  -  found in `src\ResponseEngine\Service\FormatterService.php:L82`


src/Translate/Helpers/MessageTemplateTranslationHelper.php
```diff
@@ -40,6 +48,334 @@ class MessageTemplateTranslationHelper
...
+        );
+        $message->setTitle($title);
+        $subtitle = TranslateService::getMessageTranslationString(
+            $messageUid,
+            OpenDialogMessage::RICH . '-' . OpenDialogMessage::SUBTITLE . '-' . $messageCount,
+            $scenarioUid,
+            $targetLanguage
+        );
+        $message->setSubTitle($subtitle);
+
```
goranambar, line 186: Missing the image translation here.

src/Translate/Tests/MessageTemplateTranslationHelperTest.php
```diff
@@ -8,10 +8,1197 @@
...
-    public function testAddTextMessageTranslation()
+    public function testTranslateTextMessage(): void
+    {
+        list($scenario, $uid, $language) = $this->getScenarioUidAndLanguage();
+
+        $message = new WebchatTextMessage();
+        $initialText = 'This is a test, this is only a test.';
+        $message->setText($initialText);
+
+        $translation = TranslateService::addOrUpdateMessageTranslation(
```
goranambar, line 44: `$translation` is actually the `Translation Id`, with which we started, but not using it anymore. So instead of using this value below, you can just update the message with some random string, like:
`// Update the message to the new translation string
 $message->setText('Some string');`
 Having said that, the tests are not wrong, but `TranslateService::addOrUpdateMessageTranslation` is not meant to be used and may be dropped in the future.

src/Translate/Tests/MessageTemplateTranslationHelperTest.php
```diff
@@ -8,10 +8,1197 @@
...
+    }
+
+    public function testTranslateMetaMessage(): void
+    {
+        list($scenario, $uid, $language) = $this->getScenarioUidAndLanguage();
+
+        $initialText = 'This is a test, this is only a test.';
+        $message = new WebchatMetaMessage();
+
+        $translation = TranslateService::addOrUpdateMessageTranslation(
```
goranambar, line 75: I think it would have been much easier for you if you've had just used `MessageTemplateTranslationHelper` here instead of adding/formatting the translate strings yourself. For example here, you could have used the example from `testAddMetaMessageTranslation` 
`$markup = '<message disable_text="1"><meta-message><data name="progressText">myValue</data></meta-message></message>';
        $messageTemplate = new MessageTemplate();
        $messageTemplate->setUid('uid');
        $messageTemplate->setMessageMarkup($markup);

        MessageTemplateTranslationHelper::addOrUpdateMessageTemplateTranslations($messageTemplate, 123, 'en');`
        And then work from here with your helper function.
        Because this way, if we change the add message signature, we would have to update all of the get translation tests also.

goranambar (changes requested):
@oliversarfas this looks good. Apart from couple of comments I dropped below, I would really like that the tests use `addOrUpdateMessageTemplateTranslations` and not you constructing the id's yourself, so we can be sure that your `getWebchatMessageTranslations` will match that. I dropped you a comment below about this. And also, I think the tests will be cleaner and require less code like this.

goranambar:
You can even use the existing `testAdd...` functions and just `assert` that everything is returned OK. As per my [commit](https://github.com/opendialogai/core/pull/719/commits/a632ae96b6d664047858630855a526116be70a15).

goranambar:
Hey @stu-greenshoots, this got more larger then I expected. Couple of notes:

1. `MessageTemplateTranslationHelper` got very big. So I extracted translators in separate classes. I think it is more maintainable this way. If we add new message, we'll have to add new Translator and just attach it to the `translators` array in `MessageTemplateTranslationHelper`.
2. I've updated the tests, so I'm testing `addOrUpdateMessageTemplateTranslations` with `getWebchatMessageTranslations` (unlucky choice for a name, I know, but I just continued on this) and then updating translations directly and then again pulling `getWebchatMessageTranslations` for a `MessageTemplate` and checking if they are correct.
3. I've updated the `FormMessage` also. When I was adding the `fromXml` method, I was following `WebchatMessageFormatter` and it was parsing the elements as arrays. But, `getMarkUp` requires the elements to have `getMarkUp` functions and be objects, so since we're now using `OpenDialogMessages::getMarkUp`, I needed to update that,
4. `AutoCompleteMessage` had invalid `XML` created in `getMarkUp()` function. Not sure how was this used until now?
5. I've added attributes to `OpenDialogMessages` class, so we can recreate the `XML` on `getMarkUp`.

stu-greenshoots (approved):
Happy with this, thanks @goranambar . I made a small commit to add methods for generating key names if you could take a look at that

PR #720, opened 2023-03-01
Add IntentClassifierConfiguration (with interfaces)

andrii-movadex:


src/LanguageProcessorEngine/Configuration/BaseLanguageProcessorConfiguration.php
```diff
@@ -7,11 +7,30 @@
...
         return new ComponentConfigurationKeyWithName($this->configurationName);
     }
+
+    /**
+     * @inheritDoc
+     */
+    public function setType(string $type): void
+    {
+        $this->type = $type;
+    }
```
elliotmassen, line 26: Let's remove this setter method. Please see `InterpreterWithEntitiesConfiguration` for an example of the pattern we use for setting properties in configurations. We will expect the `type` data to be present in the `$data` parameter of the constructor of this configuration (stored in the `configuration` field on the `ComponentConfiguration` model).

src/LanguageProcessorEngine/Tests/Configuration/ConfigurationTest.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace OpenDialogAi\LanguageProcessorEngine\Tests\Configuration;
+
+use OpenDialogAi\Core\Components\Configuration\ComponentConfigurationKeyWithName;
+use OpenDialogAi\Core\Tests\TestCase;
+use OpenDialogAi\LanguageProcessorEngine\Configuration\BaseLanguageProcessorConfiguration;
+
+class ConfigurationTest extends TestCase
```
elliotmassen, line 9: Let's add a test here to check that the `type` property is successfully get/set.

elliotmassen (changes requested):

src/LanguageProcessorEngine/Configuration/BaseLanguageProcessorConfiguration.php
```diff
@@ -7,11 +7,30 @@
 
 class BaseLanguageProcessorConfiguration extends BaseConfiguration implements LanguageProcessorConfigurationInterface
 {
+    const TYPE_KEY = 'type';
```
elliotmassen, line 10: Let's just call this TYPE for simplicity.

```suggestion
    const TYPE = 'type';
```

elliotmassen (changes requested):
Thanks Andrii! Just two final small tweaks to make.

src/LanguageProcessorEngine/Configuration/BaseLanguageProcessorConfiguration.php
```diff
@@ -7,14 +7,15 @@
 
 class BaseLanguageProcessorConfiguration extends BaseConfiguration implements LanguageProcessorConfigurationInterface
 {
-    const TYPE_KEY = 'type';
+    const TYPE = 'type';
     private string $type;
+    protected static array $requiredAttributes = [self::TYPE];
```
elliotmassen, line 12: Apologies this is my bad for not being specific enough, this wants to be the `$required` property from `BaseConfiguration`.

```suggestion
    protected static array $required = [self::TYPE];
```

src/LanguageProcessorEngine/Tests/Configuration/ConfigurationTest.php
```diff
@@ -43,9 +44,15 @@ public function testSetupOfConfigurationTypeField()
...
-        $configuration = BaseLanguageProcessorConfiguration::create('my_config', [BaseLanguageProcessorConfiguration::TYPE_KEY => $configurationType]);
+        $configuration = BaseLanguageProcessorConfiguration::create('my_config', [BaseLanguageProcessorConfiguration::TYPE => $configurationType]);
 
         // Assert
         $this->assertEquals($configurationType, $configuration->getType());
     }
+
+    public function testConfigurationCreationWithoutTypeField()
+    {
+        $this->expectException(ErrorException::class);
```
elliotmassen, line 55: I think we'd expect `InvalidConfigurationDataException` here instead.

elliotmassen (changes requested):

src/LanguageProcessorEngine/Configuration/BaseLanguageProcessorConfiguration.php
andrii-movadex: No worries, I also could be more attentive. Fixed.

elliotmassen (approved):

PR #721, opened 2023-03-01
ODTR 62 | CSV file import and command

goranambar:
Adds import command for translation `csv` files in `TranslateService` + adds command `scenario:import-translation` for importing translation file.


src/Translate/Services/Formatters/CsvTranslationFormatter.php
```diff
@@ -42,4 +52,94 @@ public static function export(string $baseLanguage, array $targetLanguages, Coll
...
+            $data = $line;
+            $group = $data[0];
+            $key = $data[1];
+            $baseTranslation = $data[self::BASE_LANGUAGE_INDEX];
+            $targetTranslations = array_slice($data, self::TARGET_LANGUAGE_INDEX);
+            $languageLine = OpenDialogLanguageLine::where('group', $group)
+                ->where('key', $key)
+                ->first();
+            if (!$languageLine) {
+                // What to do here? What if, for example, Message is deleted in the meantime?
```
stu-greenshoots, line 76: I think this is the right thing to do in this case, log the warning and move on. Since the message ID is used in the keys, it would never be valid

src/Translate/Services/Formatters/CsvTranslationFormatter.php
```diff
@@ -42,4 +52,94 @@ public static function export(string $baseLanguage, array $targetLanguages, Coll
...
+                Log::warning('Language line not found: ' . $group . '.' . $key . ' Skipping import...');
+                continue;
+            }
+            $languageLine->setTranslation($baseLanguage, $baseTranslation);
```
stu-greenshoots, line 80: I think the business logic says that the base language doesn't get updated, only the target languages- but let's check this

stu-greenshoots (approved):

stu-greenshoots (changes requested):
There's a small change needed to the logic for the base translation here I think

PR #722, opened 2023-03-02
Create IntentClassifier (with interfaces)

andrii-movadex:

PR #723, opened 2023-03-02
Setup configurations

andrii-movadex:


elliotmassen (changes requested):
Thanks Andrii! Would you be able to ensure this file is consistent with the naming used for the file name & array keys found in other engines? (see Interpreter Engine or Action Engine config files for example)

elliotmassen (approved):
Thanks for making those changes! This looks good to me 👍  I'll get this merged down once https://github.com/opendialogai/core/pull/720 & https://github.com/opendialogai/core/pull/722 are approved too.

PR #724, opened 2023-03-02
Setup Language Service Provider

andrii-movadex:


elliotmassen (changes requested):
Thanks for this Andrii! Really great to see the test clean up here too - thank you 👍 Please see my comments below:

- [ ] Would you be able to update the class doc block comment for `ComponentConfiguration` to include the new scope method please?

elliotmassen (approved):

PR #725, opened 2023-03-06
AutoTranslate Interfaces

stu-greenshoots:
This PR adds in the interface and base classes required for the auto-translate functionality along with some tests. 

This is a scaffolding PR that will be built upon to build up the functionality


goranambar (changes requested):
I'm not sure what should we do and how we should handle sentences with emojis? For example:
💡 For example : 2 mg per day

PR #726, opened 2023-03-07
Create types that define the appropriate data model

andrii-movadex:


src/LanguageProcessorEngine/Model/Dataset.php
```diff
@@ -0,0 +1,53 @@
...
+    ];
+
+    /**
+     * Gets related ComponentConfiguration model
+     *
+     * @return BelongsTo
+     */
+    public function componentConfiguration() : BelongsTo
+    {
+        return $this->belongsTo(ComponentConfiguration::class);
```
elliotmassen, line 27: I wonder whether this PR should also include the inverse change to `ComponentConfiguration` to be able to optionally call query the dataset - do you think that's worth adding here?

elliotmassen (changes requested):
This looks good 👍 Just a few comments below:

 - [x] Let's move the `$casts` arrays to be just under the `$fillable` declarations.
 - [x] Could you please add class doc comments to the models to document their properties?
 - [x] Please prefix all the migration's table names with `language_processor_`. This will ensure we don't create any confusion with other internal uses of intents or utterances.

src/LanguageProcessorEngine/Model/Dataset.php
andrii-movadex: According to our communication in slack, we can leave this as it is for now. May be later if it will be required and we will add an inverse relation method. Hope it makes sense. Thanks

elliotmassen (changes requested):
Thanks for making these changes! The migration table name changes are correct - however they are missing the respective updates to the table names referenced in the `down` methods.

andrii-movadex:
> Thanks for making these changes! The migration table name changes are correct - however they are missing the respective updates to the table names referenced in the down methods.

Sorry for that. Fixed

elliotmassen (approved):

PR #727, opened 2023-03-07
Connect new Language service engine to the SAAS app via Reflection

andrii-movadex:
…lving of LanguageProcessorEngineReflectionInterface (inside ReflectionServiceProvider); ADDED resolving of LanguageProcessorEngineReflectionInterface inside ReflectionHelper; ADDED test.


src/Reflection/Reflections/LanguageProcessorEngineReflection.php
```diff
@@ -0,0 +1,86 @@
...
+     * @inheritDoc
+     */
+    public function jsonSerialize()
+    {
+        $languageProcessors = $this->getAvailableLanguageProcessors();
+
+        $languageProcessorsWithData = array_map(function ($languageProcessor) {
+            return [
+                'component_data' => (array) $languageProcessor::getComponentData(),
+                'language_processor_data' => [
```
elliotmassen, line 55: Let's also add in a `types` field here that specifies the types this processor provides.

src/Reflection/Reflections/LanguageProcessorEngineReflection.php
```diff
@@ -0,0 +1,86 @@
...
+                ],
+            ];
+        }, $configurations->values()->toArray());
+
+        $configurationsWithData = array_combine($configurations->keys()->toArray(), $configurationsWithData);
+
+        return [
+            "available_language_processors" => $languageProcessorsWithData,
+            "available_configured_language_processors" => $configurationsWithData
+        ];
```
elliotmassen, line 84: I think it would be worth adding a `engine_configuration` field here (see how we do this in `InterpreterEngineReflection` for an example). This engine configuration should iterate through each of the registered processor components, create an array of all their provided types (via `getTypes`), and then return an array of unique types. This will provide the front-end with a unique list of types to offer to the user.

elliotmassen (changes requested):
Thanks Andrii, this is really great! This will also require a sibling PR in the application in order to return this data from the reflection endpoint.

Would you also be able to add a test like `testGetAvailableConfiguredLanguageProcessors` to test this aspect of the reflection too?

andrii-movadex:
> Thanks Andrii, this is really great! This will also require a sibling PR in the application in order to return this data from the reflection endpoint.
> 
> Would you also be able to add a test like `testGetAvailableConfiguredLanguageProcessors` to test this aspect of the reflection too?

I've created a PR https://github.com/movadex/opendialog-saas/pull/3 for this task in saas project. Please check.
Also I've added mentioned test to core project. Please check.

PR #728, opened 2023-03-08
AWS Translator service

stu-greenshoots:
This is a draft PR that implements the AWS translator through the use of an SDK.

This is more a proof of concept until it has been readied for production:

- More tests
- Use of response classes
- Better Error handling


goranambar (approved):
This looks good @stu-greenshoots I've added some updates so if you're happy with them go ahead and merge.
I've tried it locally and managed to get translations in `mk` and `es`. Looks good!

PR #730, opened 2023-03-09
ODTR-93 | API resource for counting translations

goranambar:
Updates `getTranslationLanguagesByScenario` to return count of total/translated strings.


stu-greenshoots (approved):
Thanks Goran, looks good and works as expected

PR #732, opened 2023-03-13
ODTR-104 | Do not store empty translation strings, delete historic ones where appropriate

oliversarfas:
Updates the TranslateService to delete any existing empty strings (where trying to update) and do not store new empty strings.

Tests updated to reflect new expected functionality


goranambar (approved):
@oliversarfas I've just made a little update to your code: even if we are not forcing update we should delete empty language line entry. Otherwise, it looks OK. I've tested locally by importing `MMM` scenario and I can see that empty translatable content isn't imported.

PR #739, opened 2023-03-16
ODTR-102 | Default Language for Scenario Imports

oliversarfas:
Adds class constant for default language imports when using flat file import


goranambar (approved):
Looks good!

PR #742, opened 2023-03-17
Fix BaseTranslationFormatter

stu-greenshoots:
This PR fixes the character that the key is split on from `.` to `-` within the BaseTranslationFormatter. This was posing a problem on postgres, but not mysql


goranambar (approved):
lgtm

PR #747, opened 2023-03-21
Fix relations in migration files for language_processor_ tables

andrii-movadex:


elliotmassen (approved):
Just a note to say that I've checked the latest release and it seems the original versions of these migrations from https://github.com/opendialogai/core/pull/726 haven't been released yet - so I'm happy enough to merge these changes as they shouldn't have been migrated on a remote environment yet.

PR #749, opened 2023-03-21
Escape CSV characters

stu-greenshoots:
This PR updates the csv formatter to use built-in CSV writing methods to ensure characters are properly escaped.

It makes use of fputcsv to write to a temporary file and then read it as a stream and return the string


goranambar (approved):
I've tried it locally and generated translation export for Side Effect reporting scenario and the export looks ok.

PR #750, opened 2023-03-21
Auto translate languages to translate

stu-greenshoots:
This PR updates the translate service to take a list of languages to be auto-translated, or to default to the list of target languages if one is not provided


src/Translate/Services/TranslateService.php
```diff
@@ -160,21 +160,26 @@ public function exportTranslationFile(
...
-        bool $autoTranslate = false
+        bool $autoTranslate = false,
+        array $languagesToBeAutoTranslated = []
     ): string {
         $languageLines = OpenDialogLanguageLine::where('scenario_id', $scenarioId)->get();
 
         if ($this->allowAutoTranslate && $autoTranslate) {
             try {
-                $languageLines = $this->translateLanguageLines($languageLines, $baseLanguage, $targetLanguages);
+                if (empty($languagesToBeAutoTranslated)) {
```
goranambar, line 170: Probably this should not happen, but do we want to this: translate to all target languages? Just contemplating, not sure.

goranambar (approved):
Ok, this looks good, apart from that comment about falling back to translate all of the target languages.

src/Console/Commands/ScenarioExportTranslation.php
```diff
@@ -20,7 +21,9 @@ class ScenarioExportTranslation extends Command
      *
      * @var string
      */
-    protected $signature = 'scenario:export-translation {fileType} {baseLanguage} {scenarioId} {targetLanguages*}';
+    protected $signature =
+        'scenario:export-translation
+        {fileType} {baseLanguage} {scenarioId} {targetLanguages} {languagesToTranslate} {--autoTranslate=false}';
```
goranambar, line 26: Can we have `languagesToTranslate` to be an option also? Because if `autoTranslate` is false we actually don't need that.

src/Console/Commands/ScenarioExportTranslation.php
```diff
@@ -41,7 +44,10 @@ public function handle()
         $fileType = $this->argument('fileType');
         $baseLanguage = $this->argument('baseLanguage');
         $targetLanguages = explode(',', $this->argument('targetLanguages'));
+        $languagesToTranslate = explode(',', $this->argument('languagesToTranslate'));
```
goranambar, line 47: The same as before: `languagesToTranslate` can be an option and only checked if `autoTranslate` is `true`.

goranambar (changes requested):
Looks good. Just I think that `languagesToTranslate` should be an option too, because we are not going to need that if `autoTranslate` is `false`.

goranambar (approved):
Looks good!

PR #754, opened 2023-03-23
Fix form message translation

stu-greenshoots:
This PR adds a number element class for form message parsing so that number elements are considered.

It also creates a new BaseMessageTranslator abstract class containing shared methods including a new one that fetches translations and accounts for the TranslationNotFoundException to prevent breaking a message translation as was seen when trying to translate a form message without a cancel text


stu-greenshoots:
Closing as merged into the options message PR

PR #757, opened 2023-03-31
Delete single language 

stu-greenshoots:
This PR updates the translate service to add a new method that removes a single language from the translations for a single scenario.

It also adds a command that can be used to interact with the method.


goranambar (approved):
Thanks @stu-greenshoots , looks good.

Have tried it locally and I can see that the scenario delete translation command deletes the required language.

PR #759, opened 2023-04-05
PHP8.1

stu-greenshoots:
This PR updates core to work with PHP 8.1 as a minimum by updating its composer dependency.

It also updates the CircleCI config to use a php8.1 image to run tests.

1 test had to be updated to fix the type, presumably this was a stricter check in the newer PHP


patshone-gsl (approved):
Tested locally

PR #761, opened 2023-04-06
ODTR-132 | Message translation fails if the message template does not have a scenario ID

goranambar:
Checks if message template contains underlying scenario during translation. If there's no attached scenario to it, it won't attempt translation.


stu-greenshoots (approved):
Thanks, looks good 👍

PR #765, opened 2023-04-18
Connect entity with part of training phrase (label/unlabel)

andrii-movadex:
… to language_processor_annotations; ADDED migration to add utterance_part field to language_processor_annotations table; RENAMED UtteranceEntity model to Annotation; ADDED utterance_part field to Annotation model; UPDATED table name for Annotation model; RENAMED utteranceEntities relation with annotations name (inside Entity, EntityComponent, Intent and Utterance models).


elliotmassen (changes requested):
Thanks Andrii! I think we will need a parallel PR in the application to update tests, etc to use the renamed model and the new property.

elliotmassen (approved):

PR #767, opened 2023-04-21
Do not translate default value in forms

stu-greenshoots:
This PR removes default_value from being considered in translations so that they are not included in translation files, and if they are already in the languauge_lines table, they will not be considered when translating a message.

This is because the value needs to map back to a value in the form, and we do not allow translation of form values - only of form keys


goranambar (approved):
Looks good.

PR #768, opened 2023-04-27
Stored event cache clear

elliotmassen:
This PR ensures that stored event cache data is unset, and also adds an optional chunking option that allows data persistence to be chunked, with each chunk being unset afterwards.


stu-greenshoots:
This has been superseded by the work to improve analyse and will be closed

PR #773, opened 2023-04-28
Investigating algorithm to find intents based on user utterances

andrii-movadex:
…logLanguageProcessor class.


src/LanguageProcessorEngine/Component/OpenDialogLanguageProcessor.php
```diff
@@ -23,7 +28,40 @@ public function __construct(LanguageProcessorConfigurationInterface $configurati
...
+            ]
+        )->sortBy([['confidence', 'desc']])->first();
+
+        /** @var IntentCollection */
+        $intentCollection = new IntentCollection();
+
+        if ($mostSimilarUtteranceAndItsConfidence) {
+            /** @var Intent|null */
+            $intent = Intent::whereHas('utterances', function ($query) use ($mostSimilarUtteranceAndItsConfidence) {
+                $query->where('utterance', $mostSimilarUtteranceAndItsConfidence['utterance']);
```
elliotmassen, line 54: Let's update this to use the utterance's ID as this will be more performant as a database query.

src/LanguageProcessorEngine/Component/OpenDialogLanguageProcessor.php
```diff
@@ -23,7 +28,40 @@ public function __construct(LanguageProcessorConfigurationInterface $configurati
      */
     public function processIntentClassifier(UtteranceAttribute $utterance): LanguageProcessorResult
     {
-        return new LanguageProcessorResult(new IntentCollection());
+        /** @var SimilarityServiceInterface */
+        $similarityService = new SimilarityService();
+
+        /** @var string */
+        $usersInput = $utterance->getText();
```
elliotmassen, line 35: We will require a conditional here to check whether the utterance text is empty or null, and if so we should return here an empty set of intents.

src/LanguageProcessorEngine/Component/OpenDialogLanguageProcessor.php
```diff
@@ -23,7 +28,40 @@ public function __construct(LanguageProcessorConfigurationInterface $configurati
...
+
+        /** @var Collection */
+        $utterancesFields = Utterance::where('utterance', 'LIKE', '%' . $usersInput . '%')->pluck('utterance');
```
elliotmassen, line 38: We will need to only select from utterances that belong to a given dataset.

Also, I wonder if the use of `LIKE` here might be constraining things a little too much here. For now let's grab all utterances in the dataset.

elliotmassen (approved):

elliotmassen (approved):

PR #776, opened 2023-05-14
Reflection for whether or not an interpreter is a language service interpreter

andrii-movadex:
…f configured interpreters; UPDATED tests.


src/Reflection/Reflections/InterpreterEngineReflection.php
```diff
@@ -80,14 +81,18 @@ public function jsonSerialize()
...
             return [
                 'component_data' => (array) new OpenDialogComponentData(
                     ODComponentTypes::INTERPRETER_CONFIGURATION_TYPE,
                     $interpreterConfiguration->getConfigurationSource(),
                     $interpreterConfiguration->getConfigurationName()
                 ),
                 'configured_interpreter_data' => [
-                    'component_id' => $configuredInterpreter::getComponentId(),
+                    'component_id' => $componentId,
+                    'is_language_service_interpreter' => $componentId === LanguageProcessorInterpreter::getComponentId(),
```
elliotmassen, line 95: Let's move this within the base interpreter configuration so that it is nested within the `configuration` object. 

I think we may want to rename this field something like `is_intent_classifier_language_service` and derive it from being a Language Processor interpreter that is connected to an `intent_classifier` type processor.

elliotmassen (approved):

PR #777, opened 2023-05-15
Fix/opndlg 1936

jacobrstone:
- split up bulk upsert to use chunking so that every attribute is not persisted all at once 
- removed manual persist calls throughout the user context
- tests currently pass and from my minimal user testing everything seems OK.


src/ContextEngine/DataClients/UserAttributeContextClient.php
```diff
@@ -72,6 +72,19 @@ public function persistAttributes(string $contextName, string $userId, Attribute
...
+        $batchSize = 50;
+
+        if (is_null($batchSize) || $batchSize >= count($insertData)) {
+            UserAttributes::upsert($insertData, ['user_id', 'attribute'], ['value']);
+            unset($insertData);
+        } else {
+            $chunks = array_chunk($insertData, $batchSize);
+            foreach ($chunks as $chunk) {
+                UserAttributes::upsert($chunk, ['user_id', 'attribute'], ['value']);
+                unset($chunk);
```
stu-greenshoots, line 84: $chunk here will be overwritten on the next loop.
I think we're better off unsetting $chunks and $chunk outside of the loop to make sure they are cleaned up

src/ContextEngine/DataClients/UserAttributeContextClient.php
```diff
@@ -72,6 +72,19 @@ public function persistAttributes(string $contextName, string $userId, Attribute
             ]);
         }
 
-        UserAttributes::upsert($insertData, ['user_id', 'attribute'], ['value']);
+        $batchSize = 50;
```
stu-greenshoots, line 75: Could you make this a property of the class so it's easier to find and edit? Could you also bring it down to batches of 10 as

stu-greenshoots (changes requested):
Thanks, @jacobrstone , I've tested this and it works as expected. 
However, I want to be able to do more testing on the middleware approach, but would like to get the upserting into the product. Could you please separate this into 2 branches? 1 that has the edits to the ContextClient, and 1 that adds the middleware logic. Thanks

stu-greenshoots (changes requested):
Thanks @jacobrstone , this looks good. Just 1 small comment on the placement of the comment

PR #779, opened 2023-05-22
Add option to save Entity color

andrii-movadex:
…tity model.


src/LanguageProcessorEngine/migrations/2023_05_22_094000_add_settings_field_to_language_processor_entities_table.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('language_processor_entities', function (Blueprint $table) {
+            $table->json('settings')->nullable(true);
```
elliotmassen, line 17: Was there a reason to make this an array of settings, rather than just a colour string field?

elliotmassen (changes requested):

src/LanguageProcessorEngine/migrations/2023_05_22_094000_add_settings_field_to_language_processor_entities_table.php
andrii-movadex: I've just added for future... let's say if it will be required to add some new setting. But of course we can add just color field.

elliotmassen: I'm happy with that 👍

elliotmassen (approved):

PR #782, opened 2023-05-25
Language Service fixes

elliotmassen:
This PR fixes the following aspects of the Language Service feature:

- Ensuring that Language Service configuration validation rules are applied correctly.
- Passing a Language Service's dataset ID with the component is initialised.
- Re-introducing some changes to the OpenDialog Language Service that was lost in a merge (and reworking it slightly to use the new dataset ID property).


src/LanguageProcessorEngine/Component/OpenDialogLanguageProcessor.php
```diff
@@ -28,40 +29,45 @@ public function __construct(LanguageProcessorConfigurationInterface $configurati
      */
     public function processIntentClassifier(UtteranceAttribute $utterance): LanguageProcessorResponse
     {
-        /** @var SimilarityServiceInterface */
+        $datasetId = $this->configuration->getDatasetId();
+
+        if (is_null($datasetId)) {
+            return new LanguageProcessorResponse(new IntentCollection());
+        }
+
```
andrii-movadex, line 37: May be we should not have this check because there is no sense to use language processor without dataset... So may be some code should be refactored on Saas project to create language processor together with dataset inside eloquent transaction (I know that I added DatasetSubscriber on saas project but may be that was not the best idea..)

src/LanguageProcessorEngine/Component/OpenDialogLanguageProcessor.php
```diff
@@ -28,40 +29,45 @@ public function __construct(LanguageProcessorConfigurationInterface $configurati
...
         /** @var string */
         $usersInput = $utterance->getText();
 
-        /** @var Collection */
-        $utterancesFields = Utterance::where('utterance', 'LIKE', '%' . $usersInput . '%')->pluck('utterance');
+        $utterances = Dataset::query()
+            ->with('intents', 'intents.utterances')
+            ->find($datasetId)
+            ->utterances()
+            ->get();
```
andrii-movadex, line 47: Could you please clarify what is the idea to get 'intents.utterances' relation also? Will it be required for some future features? I just think that there could be many utterances etc

andrii-movadex (commented):
In general it looks good to me.
I have some questions (but they all are related more to my understanding of what is going on):
- could you please clarify what is the idea to use configuration's language_processor_name if we have component configuration's name?
- regarding getDefaultConfigurationData method: is this method used for frontend purposes to show some default information?
Thanks

src/LanguageProcessorEngine/Component/OpenDialogLanguageProcessor.php
elliotmassen: Sure, I can clarify this! My thinking is that firstly that we need to get only the utterances that are related to the dataset associated with the configuration of this component. Therefore I introduced the dataset ID property in order to query the dataset specifically. Then I figured that it's more performant to eager load the intents and their utterances, rather than make separate queries for them. Previously the query was across all utterances (from all datasets) and filtering based on a `LIKE` statement, however I think this is both too broad (as we only need utterances from one dataset), and too narrow (as the `LIKE` statement does precludes "fuzzy" matching by the similarity service). Finally it seems efficient to query the whole utterance model (as opposed to just the utterance string), as later on the in the method it requires the associated intent model, and therefore it seemed better to remove the need for this second query.

src/LanguageProcessorEngine/Component/OpenDialogLanguageProcessor.php
elliotmassen: This is a good point. I think it's potentially redundant, but I think I'd rather avoid any larger refactoring of how we create datasets right now.

elliotmassen:
Thanks for taking a look @andrii-movadex! In regard to your questions:

> * could you please clarify what is the idea to use configuration's language_processor_name if we have component configuration's name?

This is a good question, and definitely highlights an complicated area of this work. There are two configuration names that we are handling in this functionality: a language processor configuration name, and an _interpreter_ configuration name. These can be completely different, and therefore we have to store and handle these values separately.
 
> * regarding getDefaultConfigurationData method: is this method used for frontend purposes to show some default information?

Yep that's right - this method is exposed via the reflection API in order to allow the front-end to pre-populate the JSON field for a custom component. I believe this is only currently used for Actions, but it's available for Interpreters and Language Processors too for future use by the front-end.

PR #786, opened 2023-06-01
OPNDLG-1947 | Handover

goranambar:
This PR creates following classes/files:
`BaseHandover` that should contain following parameters:
`handover config` - similar to Actions's
`BaseHandoverConfiguration`

`HandoverComponentService`
`ConfiguredHandoverService`
`HandoverServiceProvider` which should be registered and read the configured custom handovers.


src/HandoverEngine/Service/HandoverService.php
```diff
@@ -0,0 +1,88 @@
...
+
+        $handover = $this->getHandoverFromMessageTemplate($messageTemplateId, $state->getScenarioId());
+        if (!$handover) {
+            return;
+        }
+
+        $result = $handover->send($attribute, $state);
+
+        if (!$result->isSuccessful()) {
+            // TODO: We should probably reset the conversational state here?
```
patshone-gsl, line 41: It would be good to resolve this one way or the other

src/Http/Controllers/WebchatIncomingController.php
```diff
@@ -100,5 +101,8 @@ public function receive(IncomingWebchatMessage $request)
 
     protected function handleUserInSystemException(UserInExternalConversationSystemException $e)
     {
```
patshone-gsl, line 103: I don't fully understand what happens here (the flow) - am assuming this means that if there's an exception thrown as part of the interaction with the external system, we catch it and further process it from here - does this just simply return the error message back to the external system?

goranambar: @patshone-gsl this is how the normal conversational flow goes (this was already implemented previously for hand to system message. I think Elliot worked on this.).
Basically, once handed over to third party system, `ConversationalState` knows that it is in handed over state and instead of continuing to process message internally in the `ConversationEngine` it stops the execution and throws `UserInExternalConversationSystemException`. So, `handleUserInSystemException` actually handles sending of the text message to the third party system.
[This is the line](https://github.com/opendialogai/core/blob/ed8c0389067ba20022180f430c4451f008cf87f4/src/ConversationEngine/ConversationEngine.php#L164) in `ConversationEngine` that does this check.

src/HandoverEngine/Service/HandoverService.php
goranambar: @patshone-gsl I've decided to remove the check for now. It will be harsh I think to get the user out of hand to system state if only one message wasn't sent. Next one might come through.
If there's any error, it will be logged in the handover implementation, so we'll have logs to investigate the issue.

PR #787, opened 2023-06-02
OpenAI Language Processor: Intent Classification

andrii-movadex:
…nAILanguageProcessor, OpenAILanguageProcessorConfiguration, PromptGenerator; REGISTERED OpenAILanguageProcessor in config file; ADDED OpenAILanguageProcessorTest test; FIXED testGetLanguageProcessorEngineConfiguration test.


src/LanguageProcessorEngine/Configuration/OpenAILanguageProcessorConfiguration.php
```diff
@@ -0,0 +1,60 @@
...
+    private string $apiKey;
+    private string $organization;
+    private string $model;
+
+    public function __construct(string $name, array $data = [])
+    {
+        parent::__construct($name, $data);
+
+        $this->apiKey = $data[self::API_KEY];
+        $this->organization = $data[self::ORGANIZATION];
```
elliotmassen, line 27: Thank you for updating this Andrii! There's just one final amendment to make here. The front-end is going to have a boolean `managed` field, and if it is true then we will populate the API key and organisation from config file/environment variables. If it is false then we will use the API key and organisation fields from the `$data` array. Would you be able to update the configuration (and tests) to account for this new `managed` field, and to update the logic for the API key and organisation properties?

elliotmassen (changes requested):
This all looks good Andrii - thank you for actioning the previous feedback. There's just one change to make below (which we hadn't spoken about previously).

src/LanguageProcessorEngine/config/openai-settings.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+return [
+    'api_key' => env('OPEN_AI_API_KEY'),
+    'organization' => env('OPEN_AI_ORGANIZATION'),
+    'model' => env('OPEN_AI_MODEL')
+];
```
elliotmassen, line 7: Let's put these in the Language Processor configuration file under an `openai` field.

src/LanguageProcessorEngine/Configuration/OpenAILanguageProcessorConfiguration.php
```diff
@@ -23,9 +25,9 @@ public function __construct(string $name, array $data = [])
     {
         parent::__construct($name, $data);
 
-        $this->apiKey = $data[self::API_KEY];
-        $this->organization = $data[self::ORGANIZATION];
-        $this->model = $data[self::MODEL];
+        $this->apiKey = $data[self::MANAGED] ? config(self::API_KEY) : $data[self::API_KEY];
+        $this->organization = $data[self::MANAGED] ? config(self::ORGANIZATION) : $data[self::ORGANIZATION];
+        $this->model = $data[self::MANAGED] ? config(self::MODEL) : $data[self::MODEL];
```
elliotmassen, line 30: This `model` field is not controlled by the `managed` field, and can still be read from the `$data`.

src/LanguageProcessorEngine/Tests/Component/OpenAILanguageProcessorTest.php
```diff
@@ -60,7 +59,7 @@ public function testGettingIntentsFromOpenAILanguageProcessor()
         $this->createUtterance($secondIntent->id, $thirdUtteranceName);
 
         /** @var array */
-        $defaultConfiguration = $this->getDefaultConfiguration();
+        $defaultConfiguration = $this->getConfiguration(IntentClassifierInterface::TYPE, 'test', 'test', 'test', false);
```
elliotmassen, line 62: Could you please add a test that ensures that the logic works as expected when `managed` is `true`?

elliotmassen (changes requested):
Thanks for adding this in Andrii - just three comments below for your attention.

elliotmassen (approved):

PR #790, opened 2023-06-12
Optional performance logging for Conversation Engine

elliotmassen:
This PR adds some optional logging of performance metrics to the Conversation Engine. It is resolvable against the container so that in the future there is a basis with which the implementation could be swapped with Cloudwatch or similar.


patshone-gsl (approved):
This all looks reasonable to me

PR #792, opened 2023-06-14
Adding optional shallow parameter to queries that retrieve starting conversation objects

elliotmassen:
This PR adds an optional `$shallow` parameter to queries that retrieve starting conversation objects. We have seen that some large scenarios can have an disproportionate response time for matching the welcome intent, or after a conversation transition. After some investigation it seemed that the queries related to retrieving starting conversation objects were also retrieving their parent objects, and building out a tree. As this is not required by the conversation engine, the `$shallow` parameter has been adapted to ensure that starting conversation objects are retrieved with only a single level of data (and not their parent objects).

From some local testing for matching the welcome intent among 5 starting conversations, 17 starting scenes and 65 starting turns the response time went from an average of 50 seconds down to an average of 7 seconds when using these shallow queries.

There remains inefficiencies within the queries for intents, however the conversation engine presumes that intent objects contain references to their parent and child objects, and therefore this has not been changed as this requires a more careful and considered approach. Given the decent improvement with the optimisations to starting object queries, my feeling is that we can return to the question of intent query optimisation later.


src/Conversation/DataClients/Interfaces/ConversationDataClientInterface.php
patshone-gsl, line 1: Why is the getAllStartingConversations method in the interface not updated - given that it's updated in the implementation? what about the getScenarioByUid method as well?

src/ConversationEngine/Facades/Selectors/ConversationSelector.php
patshone-gsl, line 1: I don't quite understand what's going on in the selector facades? - Why was it set to shallow already and what's the impact of the change?

patshone-gsl (approved):
Other than updates to the interface looks good to me

stu-greenshoots:
This looks good to me other than the comments already there and will prevent querying for more information than needed.

I wonder if there is any way of simplifying the code a bit with depth constraints when fetching children or descendants: https://github.com/staudenmeir/laravel-adjacency-list#depth-constraints

src/ConversationEngine/Facades/Selectors/ConversationSelector.php
elliotmassen: Just to summarise from our chat yesterday for our future reference: This was a pre-existing parameter that wasn't in use, and these classes are not used our side of core. In the worst-case situation it would mean a caller who was previously incorrectly receiving a non-shallow response when asking for shallow will not get a shallow response.

PR #794, opened 2023-06-22
Links are not hyperlinked (only plain text)

branko-stankovski:
This PR adds a check for link tags in message templates and formats the links after the translation if they are converted into HTML hyperlinks.


goranambar:
The work looks nice @branko-stankovski and I tried it locally and it fixes the issue, but I think we have a deeper problem here to address. Maybe @stu-greenshoots we can catch-up on this on Monday?

Let me explain where I think that the problem lies:
1. This problem only manifests itself if we have a `link` into the label, text etc.

2. If we take for example the `OptionsMessage` we can [see here](https://github.com/opendialogai/core/blob/85c3cd0e7f1eee6e55a793cd44ee16770ed1560a/src/MessageBuilder/Message/OptionsMessage.php#L77) how it handles the `label`. What `getMessageText` actually does is transform `<link>` into html `hyperlink`, [here](https://github.com/opendialogai/core/blob/85c3cd0e7f1eee6e55a793cd44ee16770ed1560a/src/MessageBuilder/Message/OpenDialogMessage.php#L114). So, in the translation file we get `<a>` instead of our xml description `<link>`.

3. After that, when the message go through the `WebchatMessageFormatter` [here](https://github.com/opendialogai/core/blob/85c3cd0e7f1eee6e55a793cd44ee16770ed1560a/src/ResponseEngine/Formatters/Webchat/WebChatMessageFormatter.php#L534), there's no `<link>` but `<a>`, so they are already pre-formatted, but the `WebchatMessageFormatter` does not understand this.

4. What Branko did is, after messages run through translations, he used `DialogflowMessageTransformer::formatLinks` to format `<a>` back to `<link>`, which `WebChatMessageFormatter` understands.

5. The proper thing to do here is to probably update `OpenDialogMessage::getMessageText` here, to return `xml` text, instead of formatting it to a `hyperlink`. This will mean that in translation file, instead of:
```
I agree to your <a class=\"linkified\" target=\"_blank\" href=\"https:\/\/www.bayer.com\/en\/conditions-of-use\">Conditions of Use<\/a>
```
we will get:
```
I agree to your
        <link new_tab="true">
          <text>Conditions of Use</text>
          <url>https://www.bayer.com/en/conditions-of-use</url>
        </link>
```
Not sure if we make this change how will this affect MMM, because they already have existing translations?

Anyway, worth chatting through on Monday, I think.

stu-greenshoots:
@goranambar - sorry I haven't had time to respond to you. 2 things to mention:

- I have pulled this ticket in as is to MMM project as it fixes the issue.
- I agree that as we have already done translations, I'm not sure on the impact for MMM. 

Let's have a chat offline about this and work out the bets way forward

PR #795, opened 2023-06-27
Feature/persist attributes middleware

jacobrstone:
This PR adds some tests for the PersistAttributesMiddleware to ensure the middleware is called, and the middleware persists the data from the user contexts to the user_attributes table in the tenant DB.


stu-greenshoots (changes requested):
Thanks for this Jacob, the code looks good and is doing as expected. 

A couple of points where we have lines that can be removed as they are not needed. I've also spotted another place explicitly calling the persist method - in the action performer. 

Both here and the expected attribute handler, we are actually calling persist on all persistent contexts, which makes me think we want to ensure all of these are caught in the middleware. We could do this by looping through all contexts from `ContextService::getContexts()` and calling persist on any that are instances of `PersistentContext`

Lastly, I think it could be worth making this a terminable middleware so that the user response is not affected by time to persist

stu-greenshoots (approved):
Thanks @jacobrstone , looks good. I have made an update to increase batch size and remove the call to persist in the action performer

PR #798, opened 2023-07-03
Create the webhook interpreter

branko-stankovski:
This PR create the webhook interpreter which returns a collection of intents (one intent in this case that has the biggest confidence value based on the response data received from the client).

The interpreter returns an Intent with a confidence score, intent name, OD ID and the associated attributes.

The URL of the external service and the headers for the request are stored in the webhook interpreter configuration.


src/InterpreterEngine/Webhook/WebhookResponse.php
```diff
@@ -0,0 +1,64 @@
...
+            Log::warning("Retrieving response data failed", [
+                'Exception message' => $e->getMessage()
+            ]);
+        }
+    }
+
+    /**
+     * @return WebhookIntent|null
+     */
+    public function getTopScoringIntent(): ?WebhookIntent
```
stu-greenshoots, line 28: As discussed, I don't think we need this as the conversation engine will compare all intents and their relative scores to the available upcoming intents

src/InterpreterEngine/Interpreters/WebhookInterpreter.php
```diff
@@ -0,0 +1,110 @@
...
+        return $collection;
+    }
+
+    /**
+     * @param WebhookResponse $response
+     * @return Intent|null
+     */
+    protected function createODIntent(WebhookResponse $response): ?Intent
+    {
+        $intent = $response->getTopScoringIntent();
```
stu-greenshoots, line 80: Here, we should just return all intents in the response

stu-greenshoots (changes requested):
Thanks Branko, just a couple of notes to action on the top scoring intent logic.

Also 1 small thing if you could look into it. When selecting the custom interpreter through the front end, the configuration is not pre-populated as it is for actions:
image

vs

image

It seems like this could be a general issue as it is the same for other interpreters locally for me

branko-stankovski:
> Thanks Branko, just a couple of notes to action on the top scoring intent logic.
> 
> Also 1 small thing if you could look into it. When selecting the custom interpreter through the front end, the configuration is not pre-populated as it is for actions: image
> 
> vs
> image
> 
> It seems like this could be a general issue as it is the same for other interpreters locally for me
> 

Since this is a FE issue, the fix is included in this PR: https://github.com/opendialogai/opendialog-design-system/pull/492 in this commit: 8e78323d25e6e4fc88c03945e64daa8605db7ff3

src/InterpreterEngine/Webhook/WebhookResponse.php
branko-stankovski: Makes sense. I removed the getTopScoringIntent() method.

src/InterpreterEngine/Interpreters/WebhookInterpreter.php
branko-stankovski: I updated the method to return all intent. Also, the method name is changed from createODIntent() to createODIntents().

src/InterpreterEngine/Webhook/WebhookIntent.php
```diff
@@ -0,0 +1,64 @@
...
+    public const WEBHOOK_INTENT = "intent.core.WebhookIntent";
+
+    protected array $entities = [];
+
+    public function __construct(string $name, float $confidence, array $entities = [])
+    {
+        parent::__construct();
+        parent::setConfidence($confidence);
+
+        $this->setODId(self::WEBHOOK_INTENT);
```
stu-greenshoots, line 19: Sorry @branko-stankovski , I just noticed that this is not quite right. The OD ID is used to match the intent, so it should be assigned the name that is returned from the API call dynamically.

stu-greenshoots (changes requested):
1 small changed needed now that I can see it flowing through the test window

src/InterpreterEngine/Webhook/WebhookIntent.php
branko-stankovski: @stu-greenshoots this is fixed with the last commit.

PR #799, opened 2023-07-05
Feature/opndlg 2026 state actions

stu-greenshoots:


src/AttributeEngine/config/opendialog-attributeengine.php
```diff
@@ -174,5 +174,8 @@
 
         // Translate
         'language' => StringAttribute::class,
+
+        // Actions
+        'original_state' => InferredCompositeAttribute::class
```
elliotmassen, line 179: What do you think of naming this something like `stored_state` or `stored_conversational_state`?

elliotmassen (commented):
This looks good - just a few minor comments.

PR #800, opened 2023-07-06
Meta Message translation fix

stu-greenshoots:
This PR updates the meta-message to use elements to create message XML rather than data. 

Previously, data was set in 2 different ways - in the fromXML method, it was set as raw XML. 

This change removes $data and leaves only $elements to save confusion in the future


goranambar (approved):
Ok, looks good. Tested it locally and it can translate `progressText` alone, with other attributes or without `progressText`

